PR #61, opened 2015-03-31
Fix issue 1 - Corrected parameter.yml.dist

marcoooo:
Added parameter    sylius.currency_importer.ecb.base_currency: EUR (parameter.yml.dist)
This missing parameter causes install failure.


pjedrzejewski:
Thank you Marc!

PR #63, opened 2015-05-26
Update to Sylius v0.15.0

zyleee:
- Updates Sylius to v0.15.x
- Removes FOS User Bundle dependency
- Changes Security config to use SyliusUserProvider

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,9 +22,8 @@
     "require": {
         "php":                          ">=5.3.3",
 
-        "sylius/sylius":                "0.14.*@dev",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -22,9 +22,8 @@
     "require": {
         "php":                          ">=5.3.3",
 
-        "sylius/sylius":                "0.14.*@dev",
+        "sylius/sylius":                "0.15.*@dev",
```
aRn0D: @pjedrzejewski Do we should use the stable version?

PR #65, opened 2015-06-05
Add web/media folder in deploy.rb.dist. Fixes #62

### LINKED ISSUE #62: web/media should be writable and shared in deploy.rb
Hi all,

Product images are generated on-the-fly using LiipImagineBundle inside the `web/media` folder, but this folder is not configured properly inside app/deploy.rb. It needs to be both writable (so that the images can be generated) and shared (so that the files persist from one deployment to another).

Will you accept a PR for this?

Thanks.

liviucmg:


liviucmg:
Fixes #62

pjedrzejewski:
Thank you Liviu! :+1:

PR #71, opened 2015-09-02
Session handler config fix

michalmarcinkowski:
Same as symfony-standard (https://github.com/symfony/symfony-standard/blob/2.8/app/config/config.yml#L30). 

Fixes https://github.com/Sylius/Sylius/issues/2585.


pjedrzejewski:
Thank you Michal! :+1:

PR #81, opened 2015-12-02
[Composer] Locked symfony/symfony <2.8 due to failing build

pamil:
It may help :)


michalmarcinkowski:
Not sure why PHP7 fails, since `phpcr-odm 1.3.0-rc4` is installed which should be compatible with PHP7...

PR #83, opened 2016-01-11
Updates before v0.16.0 release

pjedrzejewski:


composer.json
```diff
@@ -20,21 +20,22 @@
...
-        "behat/mink-selenium2-driver":  "^1.2",
-        "behat/mink":                   "^1.6",
-        "coduo/php-matcher":            "^1.0",
-        "phpspec/phpspec":              "^2.4",
-        "phpunit/phpunit":              "^4.1"
+        "behat/mink-selenium2-driver": "^1.2",
+        "behat/mink":  "^1.6",
+        "coduo/php-matcher": "^1.0",
+        "phpspec/phpspec": "^2.4",
+        "phpunit/phpunit": "^4.1"
```
lchrusciel: We are missing ApiTestCase and we should bump *php-matcher* to "^2.1@dev". Do we want this dependencies in sylius-standard?

pjedrzejewski: Updated. Good catch!

PR #105, opened 2016-07-28
Use FOB SuiteSettingsExtension

pamil:
It will allow to remove the extension code from Sylius main repository.


pjedrzejewski:
👍

PR #111, opened 2016-08-16
Updated security.yml to support new multi-user providers

gorkalaucirica:
Fixes issue reported in Sylius/Sylius#5796


pjedrzejewski:
Thank you very much Gorka! :)

Issue #126, opened 2016-10-13
Composer update fails


dszwcz:
The installation proceeds without a problem (because composer.lock). When I try to make composer update the following problems arise:

1. After the attempt to execute post update clearing cache command I receive a message: `[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]                       
  You have requested a non-existent service "FOSOAuthServerBundle must be registered in kernel."`

2. So I add a FOSOAuthServer bundle to AppKernel and then I see another error: `Class "model" is not configured for resource "sylius.api_user"`. ApiBundle/Resources/config/app/config.yml file should be included.

3. [NEW] In the `app/config/security.yml` config file `sylius.admin_user.provider.email_or_name_based` should be replaced by `sylius.admin_user_provider.email_or_name_based`.

[PR](https://github.com/Sylius/Sylius-Standard/pull/127) where FOSOAuthServerBundle added to the kernel.
Should I update composer.lock too?

dszwcz:
[PR](https://github.com/Sylius/Sylius-Standard/pull/127) where FOSOAuthServerBundle added to the kernel.

lchrusciel:
The second problem could be resolved by adding the following line to config file:
```yml
imports:
    - { resource: "@SyliusApiBundle/Resources/config/app/config.yml" }
```

krisElphenor:
Hello,

I did the same update and encountered the same errors, so i applied the above fixes (added FOSOAuthServerBundle to the kernel and the SyliusApiBundle config.yml to the config file).

Now, i have the following error:
`The service "security.authentication.manager" has a dependency on a non-existent service "sylius.admin_user.provider.email_or_name_based".`

Did anyone else has the same issue ? or any idea about what i did miss ?

Thanks

dszwcz:
@krisElphenor open the `app/config/security.yml` config file and change sylius.admin_user.provider.email_or_name_based to sylius.admin_user_provider.email_or_name_based

dszwcz:
Closed because all this is solved by @Zales0123

krisElphenor:
Thanks @dszwcz , i did checked that file on github, but didn't notice it has changed.

Also, i would like to point out, the next error is:
`The service "security.firewall.map.context.shop" has a dependency on a non-existent service "sylius.authenticaiton.success_handler".`

it's just a spelling mistake, it should be 
[...]

PR #128, opened 2016-10-13
composer.lock updated

dszwcz:
| Q             | A
| ------------- | ---
| Fixed tickets | [#126]
| License       | MIT


lchrusciel:
The build will be red as many things changed since last composer.lock update

Issue #123, opened 2016-10-04
Composer install ends with warning


crevillo:
Doing `composer create-project -s dev sylius/sylius-standard`, after you set your configuration params, the process ends outputting 

```
! [NOTE] Some assets were installed via copy. If you make changes to these assets you have to run this command again.  


 [OK] All assets were successfully installed.                                                                           


> Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::installRequirementsFile
Class Symfony\Cmf\Bundle\CreateBundle\Composer\ScriptHandler is not autoloadable, can not call post-install-cmd script
```

nietzcheson:
@crevillo It is not memory in your virtual machine? This happened to me. I upgraded to 4G and works well ...

crevillo:
I don't think so. But i'll try it. To me is more like a deps issue or
something. For instance, if you clone sylius master and do composer install
no warning appears.
Is only when you install sylius-standard following doc instructions.

gorkalaucirica:
This seems fixed, no cmf related warning when installing using `composer create-project -s alpha`

PR #134, opened 2016-10-29
Remove CMF commands from composer as they are not present

crevillo:
|  Q | A |
|---|---|
|  Fixed Tickets | #123 |
|  License | MIT |

Just remove the commands from the composer.json to make the warning dissapear.


pjedrzejewski:
👍 👍 👍

PR #147, opened 2016-12-02
[Travis] Enable PHP 7.1 for tests

venyii:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |

PHP 7.1 was released yesterday, so I think it's about time to run the test suite against it 🎉 

Build is failing for 7.1 because of `Warning: A non-numeric value encountered  in ...` which is fixed in https://github.com/Sylius/Sylius/pull/6757

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
 sudo: required
 
 php:
-    - 7.0
+    - 7.1
     - 5.6
 
 services:


venyii:
Replaced 7.0 with 7.1 here as well, as discussed in https://github.com/Sylius/Sylius/pull/6757

michalmarcinkowski:
Thanks! :+1:

PR #157, opened 2017-01-30
Upgrade to latest sylius + deps + fos rest bundle

bendavies:
Upgrades Sylius standard for https://github.com/Sylius/Sylius/pull/7349

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -51,6 +51,7 @@ swiftmailer:
     password: "%mailer_password%"
 
 fos_rest:
+    exception: ~
     view:
         formats:
             json: true
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -50,6 +50,3 @@ swiftmailer:
 sylius_theme:
     sources:
         test: ~
-
-twig:
-    exception_controller: "Sylius\\Bundle\\CoreBundle\\Controller\\VerboseExceptionController::showAction"
--- a/app/config/config_test_cached.yml
+++ b/app/config/config_test_cached.yml
@@ -22,3 +22,6 @@ sylius_channel:
 fos_rest:
     exception:
         debug: true
+
+doctrine_migrations:
+    dir_name: "%kernel.root_dir%/../vendor/sylius/sylius/app/migrations"
--- a/app/config/config_test_cached.yml
+++ b/app/config/config_test_cached.yml
@@ -18,3 +18,7 @@ twig:
 
 sylius_channel:
     debug: true
+
+fos_rest:
+    exception:
+        debug: true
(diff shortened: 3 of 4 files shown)


lchrusciel:
The build is failing because of missing migration from main sylius repository. We could copy and paste them from the main repository, but there is also another solution(which I found easier to maintain).
We can create a file, lets say `app/config/sylius_migration.yml`
```yml
# app/config/sylius_migration.yml

name: Sylius Migrations
migrations_namespace: Sylius\Migrations
table_name: sylius_migrations
migrations_directory: "vendor/sylius/sylius/app/migrations/"
```

and than, just run migration from this file:
```bash
$ sf d:m:m --configuration app/config/sylius_migration.yml
```
[Here](http://docs.doctrine-project.org/projects/doctrine-migrations/en/latest/reference/introduction.html#configuration) are some docs about it. 
What would you say?

bendavies:
@lchrusciel doesn't seem great. a user needs to implement their own migrations so would need more than 1 directory for migrations?

lchrusciel:
@bendavies No. the main migration folder stay the same. It just allow us, to run migration from Sylius vendor if you need. Normally, it is not taken into account. We just should put this config the content to separate file and make it easier for others to run Sylius migration without required copy-pasting.

bendavies:
@lchrusciel got ya!

bendavies:
@lchrusciel have you guys decided how you want to proceed?

lchrusciel:
It is not up to me. /cc @pjedrzejewski @michalmarcinkowski Should we copy & paste migration from main Sylius repository, or allow to use them without duplication?

michalmarcinkowski:
IMO the solution proposed by @lchrusciel is fine. @pjedrzejewski?

bendavies:
what about the existing single migration in `app/migrations` and the `doctrine_migrations` config in `config.yml`?

bendavies:
@lchrusciel i pushed a commit. same idea as yours

lchrusciel:
It is working, so good enough for me. But the question is, shouldn't we provide a way to easily run Sylius migration from Sylius-Standard... Anyway, it can be a topic for RFC

bendavies:
Yeah, i'm unsure what to do with the current `app/migrations` directory. copy the sylius ones in or remove the one that is there?

michalmarcinkowski:
How about removing the migrations from Sylius-Standard and just create a schema on travis for tests? The migrations are already tested on Sylius/Sylius repository, so there is no benefit of running them here as well.

liverbool:
Consider to use `zone` in sylius config.
http://symfony.com/doc/current/bundles/FOSRestBundle/3-listener-support.html#zone-listener

pjedrzejewski:
Thanks Ben!

PR #159, opened 2017-02-07
[Maintenance] [DX] Improve migration handling

lchrusciel:
As I mention in [this comment](https://github.com/Sylius/Sylius-Standard/pull/157#issuecomment-276076814) with this file it will be much easier to run Sylius Migrations. It would be enough to execute the following command:
```bash
$ php app/console doctrine:migration:migrate --configuration app/config/sylius_migration.yml
```
instead of copy-pasting the whole migration folder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
 before_script:
     - bin/console doctrine:database:create --env=test_cached -vvv # Have to be run with debug = true, to omit generating proxies before setting up the database
     - bin/console cache:warmup --env=test_cached --no-debug -vvv
-    - bin/console doctrine:migrations:migrate --no-interaction --env=test_cached --no-debug -vvv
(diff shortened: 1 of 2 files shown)


bendavies:
sort of done in https://github.com/Sylius/Sylius-Standard/pull/157/files#diff-b7cfc824d9ea33b675f58a6252ac56d8R26
so that needs removing if you want to do it this way

lchrusciel:
@bendavies yeah, but we receive a few questions about handling migration in Sylius - Standard on our slack. Your solution is fine for test purposes, but what if you just want to upgrade your app to newest Sylius? Just the trade-off, what will be best, this or copy-pasting.

pjedrzejewski:
I think we should copy-paste the migrations. In Sylius-Standard, you will also have your own migration. I guess part of the upgrade process will be copying the migrations. Unless we can run migrations from 2 directories, but as far as I know, not possible and can also be problematic. @lchrusciel Thoughts?

lchrusciel:
Agreed 👍

PR #161, opened 2017-02-10
[BugFix] Add getContainerLoader method to AppKernel

TheMadeleine:
It is a solution to https://github.com/Sylius/Sylius/issues/7095;

Prepared basing on this comment: https://github.com/Sylius/Sylius/issues/7095#issuecomment-268904810

Having such a `getContainerLoader` method [this guide](http://docs.sylius.org/en/latest/cookbook/checkout.html#overwrite-the-state-machine-of-checkout) works properly.

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -10,10 +10,21 @@
  */
 
 use Sylius\Bundle\CoreBundle\Application\Kernel;
+use Symfony\Component\Config\Loader\DelegatingLoader;
+use Symfony\Component\Config\Loader\LoaderResolver;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Loader\ClosureLoader;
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
1. Are we sure it has no bad side effects?
2. I think this should go to Kernel in CoreBundle.

michalmarcinkowski:
Maybe we could ask about that on Symfony Slack? We shouldn't be forced to do such thing to make the overriding work...

pjedrzejewski:
@TheMadeleine Can you post this on Symfony Slack and ask for some feedback? :)

TheMadeleine:
@pjedrzejewski Sure!

lyrixx:
Hello.

Actually this is not related to the workflow, but with the kernel + config.
And finally, Sylius does not use symfony/workflow but https://github.com/winzou/state-machine so I can not help here. 

Sorry.

PR #162, opened 2017-02-14
Use dama/doctrine-test-bundle to improve phpunit tests

dmaicher:
As also done in https://github.com/Sylius/Sylius/pull/7497/files


phpunit.xml.dist
```diff
@@ -0,0 +1,22 @@
...
+         bootstrap="app/autoload.php">
+    <testsuites>
+        <testsuite name="Sylius Standard Test Suite">
+            <directory>tests</directory>
+        </testsuite>
+    </testsuites>
+
+    <php>
+        <server name="KERNEL_CLASS_PATH" value="/app/TestAppKernel.php" />
+        <server name="IS_DOCTRINE_ORM_SUPPORTED" value="true" />
```
dmaicher: this can be removed maybe?

app/config/config_test.yml
```diff
@@ -15,7 +15,6 @@ framework:
 doctrine:
     dbal:
         dbname: "%database_name%_test"
-        driver_class: Sylius\Bundle\CoreBundle\Test\MySqlDriver
```
dmaicher: seems this is a different class than the one I deleted here: https://github.com/Sylius/Sylius/pull/7497/files#diff-89e907faa9c86340f8a0ae7df6ecdfaaL12

pamil: Sylius-Standard was upgraded before that class was renamed so it's all right :+1:

phpunit.xml.dist
pamil: It's specific to Lakion/ApiTestCase and I guess it is still required. /cc @lchrusciel

composer.json
```diff
@@ -45,7 +45,8 @@
         "phpunit/phpunit": "^5.6",
         "se/selenium-server-standalone": "^2.52",
         "sensio/generator-bundle": "^3.1",
-        "stripe/stripe-php": "^4.1"
+        "stripe/stripe-php": "^4.1",
+        "dama/doctrine-test-bundle": "^1.0"
```
pamil: It would be awesome to keep alphabetical order here 🎉

phpunit.xml.dist
lchrusciel: It has to stay, because otherwise fixtures will not be loaded.

composer.json
dmaicher: yeah just used `composer require --dev dama/doctrine-test-bundle` :yum: 

I can restore the order again :wink:

dmaicher:
@pamil I can also remove the `phpunit.xml.dist` again. I just thought it might be a nice start to have a basic config already.

dmaicher:
Changes done.

I also updated `sylius/sylius` to latest master commit.

pjedrzejewski:
Great, thank you so much David! :)

PR #184, opened 2017-08-24
[Composer] Use correct parameters.yml.dist

### LINKED ISSUE #182: Use correct parameters.yml.dist
By default we configured the parameters.yml.dist from vendors which is annoying & confusing for new projects because Composer does not ask for newly defined parameters in their own parameters.yml.dist.

lchrusciel:
It was super annoying.

Fixes #182


pjedrzejewski:
👍

PR #185, opened 2017-09-08
Get config.yml in sync with one from sylius/sylius

### LINKED ISSUE #169: User defined locale ignored during installation
When using `composer create-project -s beta sylius/standard` it asks for a locale, with the default being `en_US`, but if you define `en_GB` it gets ignored and installs `en_US` as default locale. Is there a way of preventing this other that going to `app/config/config.yml`, editing the local and running `bin/console sylius:install` again? It's quite annoying and frustrating.

ADmad:
Closes #169.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -10,11 +10,8 @@ imports:
     - { resource: "parameters.yml" }
     - { resource: "security.yml" }
 
-parameters:
-    locale: en_US
-
 framework:
-    translator: { fallbacks: ["%locale%"] }
+    translator: { fallbacks: ["%locale%", "en"] }
     secret: "%secret%"
     router:
         resource: "%kernel.root_dir%/config/routing.yml"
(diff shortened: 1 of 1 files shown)


lchrusciel:
@pamil I would merge a fix like this because locale will change on each machine and add this tip from Symfony to our deployment guide

pjedrzejewski:
Will be fixed in #187.

pjedrzejewski:
Thanks @ADmad!

ADmad:
@pjedrzejewski Well IMHO #187 didn't exactly fix the issue but rather circumvented it :slightly_smiling_face:.

I installed 1.0.0 stable and it no longer asks preferred locale to the user during installation when you run `composer create-project` as it did in beta3. So now I have to go and manually change `config.yml` before running `php bin/console sylius:install`, which new users won't be aware of.

PR #202, opened 2017-09-28
Putting back AppBundle in AppKernel

blankcheck:
Q | A
-- | --
Branch? | 1.0
Bug fix? | no
New feature? | no
BC breaks? | no
Deprecations? | no
Related tickets | -
License | MIT

Not sure if there is any reasoning behind it (I fail to see it), but in current Sylius-Standard there is no AppBundle in AppKernel.
This PR adds it back.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "require": {
         "php": "^7.1",
 
-        "sylius/sylius": "^1.0"
+        "sylius/sylius": "^1.1@dev"
     },
     "require-dev": {
         "behat/behat": "^3.2",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -22,7 +22,7 @@
     "require": {
         "php": "^7.1",
 
-        "sylius/sylius": "^1.0"
+        "sylius/sylius": "^1.1@dev"
```
pjedrzejewski: This should not be part of this PR. Could you rebase, please?

blankcheck: It wasn't initially - @pamil added these.
Anyhow, to simplify things since this is just one-liner feel free to close this PR and just add that line yourselves.
Don't get me wrong - no problem if you still want me to do it (just let me know), but it seams more efficient if you guys just put that line in next release.

thx

pamil: In fact, #204 added that (cause the change should affect 1.0 branch as well), thanks for the contribution anyway! :)

PR #213, opened 2017-11-16
[RFC] Initialize themes directory by default

pjedrzejewski:
I am working on the theme skeleton and I think it would make sense to have this directory by default.


pamil:
Thanks Paweł! 🎉

PR #230, opened 2018-05-28
Update to Sylius v1.2.0-BETA

### LINKED ISSUE #229: Symfony 4 support for v1.2.x
See related project https://github.com/Sylius/Sylius/projects/4.

pamil:
Closes #229.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,29 +21,29 @@
     ],
     "require": {
         "php": "^7.1",
-        "sylius/sylius": "1.2.*@dev"
+
+        "sylius/sylius": "1.2.*@beta",
+        "symfony/symfony": "^3.4|^4.0"
     },
     "require-dev": {
         "behat/behat": "^3.4",
-        "behat/mink": "^1.7",
+        "behat/mink": "^1.7@dev",
         "behat/mink-browserkit-driver": "^1.3",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -70,7 +70,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.1"
+            "php": "7.1.18"
```
teohhanhui: Why?

pamil: PHP 7.1.0 was not enough to install Symfony 4.1 IIRC

teohhanhui: If it's not listed in Symfony's `composer.json`, I'd think it's not important enough. And we shouldn't set the platform this way anyway. It means we risk generating a `composer.lock` that only works with the very latest patch release of PHP 4.1, which I'm sure is not your intention.

pamil: Good catch, I'll make an issue for that.

composer.json
```diff
@@ -21,29 +21,29 @@
     ],
     "require": {
         "php": "^7.1",
-        "sylius/sylius": "1.2.*@dev"
+
```
teohhanhui: Can we please stop adding this useless blank line? :rofl:

PR #232, opened 2018-06-01
Add Docker setup

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.2 (or master?)
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Sylius/Sylius#9414
| License         | MIT

TODO:

- [x] Add mailhog


app/config/config.yml
```diff
@@ -31,12 +31,12 @@ framework:
 
 doctrine:
     dbal:
-        driver: '%database_driver%'
         host: '%database_host%'
         port: '%database_port%'
         dbname: '%database_name%'
         user: '%database_user%'
         password: '%database_password%'
+        driver: pdo_mysql
```
teohhanhui: It's pointless to have database `driver` configurable via environment variable when the `server_version` is fixed here. It's also something that does not change in an application, so following [Symfony Best Practices](https://symfony.com/doc/3.4/best_practices/configuration.html#application-related-configuration):

> ### Application-Related Configuration
> > Define the application behavior related configuration options in the app/config/config.yml file.
>
> Defining these values in parameters.yml file would add an extra layer of configuration that's not needed because you don't need or want these configuration values to change on each server.

composer.json
```diff
@@ -81,7 +81,7 @@
         "symfony-var-dir": "var",
         "symfony-web-dir": "web",
         "symfony-tests-dir": "tests",
-        "symfony-assets-install": "relative",
+        "symfony-assets-install": "copy",
```
teohhanhui: This is necessary as symlinks don't work across containers.

docker-compose.yml
```diff
@@ -0,0 +1,105 @@
...
+      # Comment out the following line in production
+      - ./web:/srv/sylius/web:ro
+      # If you develop on Linux, uncomment the following line to use a bind-mounted host directory instead
+      # - ./web/media:/srv/sylius/web/media:rw
+      # If you develop on Linux, comment out the following line
+      - web-media:/srv/sylius/web/media:ro,nocopy
+    ports:
+      - "80:80"
+
+  mailhog:
```
teohhanhui: It's unfortunate that we've set

```yaml
swiftmailer:
    disable_delivery: true
```

in the `dev` and `staging` environments, where this would be useful.

So the user will have to change this value manually in the config files, if they use something like Mailhog.

SwiftmailerBundle does not support using an environment variable for `disable_delivery`, similar to https://github.com/symfony/swiftmailer-bundle/pull/159

teohhanhui:
Sure, and I'll have to do a rebase. You used the awful GitHub "update branch" button which doesn't work. :rofl:

PR #242, opened 2018-07-23
Add Docker setup

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Sylius/Sylius#9414
| License         | MIT

Replaces #232


app/config/config.yml
```diff
@@ -31,12 +31,12 @@ framework:
 
 doctrine:
     dbal:
-        driver: '%database_driver%'
         host: '%database_host%'
         port: '%database_port%'
         dbname: '%database_name%'
         user: '%database_user%'
         password: '%database_password%'
+        driver: pdo_mysql
```
pamil: Why not keep that parameter?

app/config/parameters.yml.dist
```diff
@@ -10,13 +9,13 @@ parameters:
 
     mailer_transport: '%env(SYLIUS_MAILER_TRANSPORT)%'
     mailer_host: '%env(SYLIUS_MAILER_HOST)%'
+    mailer_port: '%env(SYLIUS_MAILER_PORT)%'
```
pamil: Could we extract this change to another PR?

gulpfile.babel.js
```diff
@@ -52,6 +52,9 @@ watchShop.description = 'Watch shop asset sources and rebuild on changes.';
 export const build = gulp.parallel(buildAdmin, buildShop);
 build.description = 'Build assets.';
 
+export const watch = gulp.parallel(watchAdmin, watchShop);
+watch.description = 'Watch asset sources and rebuild on changes.';
```
pamil: Could we extract this change?

.travis.yml
```diff
@@ -85,6 +91,25 @@ script:
     - echo "Testing (Behat - brand new, javascript scenarios; @javascript && ~@todo && ~@cli)" "Sylius"
     - bin/behat --strict --no-interaction -vvv -f progress -p cached --tags="@javascript && ~@todo && ~@cli" || bin/behat --strict --no-interaction -vvv -f progress -p cached --tags="@javascript && ~@todo && ~@cli" --rerun
 
+    - docker-compose pull --ignore-pull-failures
+    - docker-compose build --pull
+    - docker-compose up -d
+    - sleep 60
```
pamil: No idea how to solve that better, just leaving a comment so it's more visible

docker-compose.yml
```diff
@@ -0,0 +1,108 @@
+version: '3.4'
```
pamil: These two different files for dev and prod would be hard to keep in sync (eg. this version differs). 

@jacquesbh mentioned this on Slack on Friday:
> yes. And it's just a step away to create a `docker-compose.traefik.yml` and explain that `docker-compose -f docker-compose.yml -f docker-compose.traefik.yml` can be use

Can we use an architecture like this for those extra containers (eg. mailhog, nodejs)?

docker/php/php-cli.dev.ini
```diff
@@ -0,0 +1,14 @@
+apc.enable_cli = 1
+date.timezone = ${PHP_DATE_TIMEZONE}
+opcache.enable_cli = 1
```
pamil: Opcache in CLI might cause some issues during development

app/config/config.yml
teohhanhui: Because it cannot be used in any meaningful way. :smile:

app/config/parameters.yml.dist
teohhanhui: But we need it here...

gulpfile.babel.js
teohhanhui: Sure. But we need it here...

docker-compose.yml
teohhanhui: It's not just about additional services...

docker/php/php-cli.dev.ini
teohhanhui: Such as? :D

nietzscheson:
The Docker configurations for the Sylius project should not be included in the standard version. The reasons that I consider main are:

- There are many out there who do not even use Docker. They do not even use virtual machines.
- The learning curve with Docker is quite complex. So its use should not be mandatory. Not all cloud payment systems allow the use of root to install certain packages on the virtual machine.
- There are other containerization projects such as:
   - Docker Container for Windows (https://github.com/docker/for-win)
   - RKT (https://github.com/rkt/rkt)

If Docker is included in the Standard version, the configurations of the Vagrant configured for Sylius are also included.

In https://github.com/Sylius/Sylius/issues/9414 @stefandoorn adds the following: "can it still be a separate repo? I prefer to have my infra stuff in separate repo, instead of combined with the application itself (SRP). Besides that, not everyone likes or is using Docker, so for these it's not needed to have it in the same repo."

Matter of which I totally agree.

So, all efforts to achieve a good configuration of Docker for Sylius (https://github.com/Sylius/docker) should focus on the repository that already exists and keep it separate.

teohhanhui:
@nietzscheson We've already had that discussion in Sylius/Sylius#9414 and on Slack. I don't see the need to bring it up again here...

nietzscheson:
So is. But this is where the code merges and I see that you want to reiterate what precisely was already discussed on the channel. I just want to make clear my position...

teohhanhui:
> how can one re-use this setup when trying to contribute to Sylius/Sylius? through copy-paste of the docker and docker compose files?

One way is to mount the host directory into, say, `/src/Sylius/Sylius`

> Also, certain services are missing

Could be added in other PR(s)

gabiudrescu:
Second thing I want to bring under your attention: running commands under the root user inside containers is going to mess up permissions on the files on the host.

Especially on development mode, I find it better to use another user inside the container that has UID and GID mapped with the one from the host. On the setup I generally use, I have the following in the docker file:

```
groupadd -g "$APPLICATION_GID" "$APPLICATION_GROUP"
useradd -u "$APPLICATION_UID" --home "/home/$APPLICATION_USER" --create-home --shell /bin/bash --no-user-group "$APPLICATION_GROUP" -g "$APPLICATION_GID"
```

This way, whenever entering the container, I make sure I use the $APPLICATION_USER and everything is OK.

Also, my understanding is that running commands inside docker containers as root may imply a security risk:
 - https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b
 - http://blog.dscpl.com.au/2015/12/don-run-as-root-inside-of-docker.html
 - https://github.com/minio/minio/issues/5742
 - https://www.reddit.com/r/docker/comments/7c51qn/stop_running_application_as_root_user_inside/
 - https://github.com/docker/compose/issues/2380#issuecomment-392530292
 - https://github.com/docker/compose/issues/1532#issuecomment-140160603
 - https://github.com/docker/compose/issues/1377 
 - https://github.com/moby/moby/issues/22415

gabiudrescu:
@teohhanhui 

```
One way is to mount the host directory into, say, /src/Sylius/Sylius
```

can you offer me an example on how one can use this docker setup to test a modification it brought to the Sylius/Sylius vendor file?

teohhanhui:
Switching user in the Dockerfile cannot work, because the Dockerfile has to be portable (and redistributable).

jacquesbh:
### User rights

For the user rights we have this:

Makefile:
```makefile
export USER_UID=$(shell id -u)
example:
	docker-compose ps
```

In our Dockerfiles:
```dockerfile
# Use www-data user
ARG USER_UID=1000
RUN usermod -u $USER_UID www-data
```

By default `www-data` has uid `1000`.
On macOS the user id is `501` usually.

This way, by using a simple variable, we avoid a lot of problems.

We don't have a makefile yet. Why not adding one? It's a good way to improve the developers flow.

### Missing containers

The missing containers are for testing purpose.

This is why I propose to use a `docker-compose.testing.yml`. We don't need to run all the containers all the time, specially Selenium and headless browsers since they are consuming a lot.

teohhanhui:
@jacquesbh Anything that involves setting uid/gid in the Dockerfile is not a solution, because of portability/redistributable

jacquesbh:
This is portable, working on Windows, Linux and macOS. We use it everyday for many projects. So…

teohhanhui:
> Missing containers

Let's keep this first PR small?

teohhanhui:
No, I'm talking about the image being portable. It should be able to be used by any user regardless of their uid/gid on the host, without having to rebuild the image.

teohhanhui:
We should support running with arbitrary uid/gid (via the `--user` flag of `docker run` and its equivalent in docker-compose.yml), but otherwise step down from root if necessary:

https://github.com/docker-library/mysql/issues/397#issuecomment-377288842
https://github.com/docker-library/php/issues/70#issuecomment-396595255

teohhanhui:
I hope we could have this PR merged soon, then others could help to make things better.

docker/php/php-cli.dev.ini
pamil: Once we had an issue when phpspec was running unexisting spec which had been removed before, the issue vanished after we disabled opcache.

app/config/config.yml
pamil: Reverted this change for now.

pamil:
Thank you, Teoh, let's make it the best Docker ever for 1.3 release 🎉

docker/php/php-cli.dev.ini
teohhanhui: To be honest, that sounds like a misconfiguration of OpCache, so we should be fine here.

PR #244, opened 2018-07-29
Tweak Travis CI config

pamil:
Changes made:
 - default test suite put into stages
 - run default stage suite without sudo
 - run Docker with sudo required
 - use mysql & memcached services only in default stage
 - use docker only in docker stage

It also removes the need for `after_success: true` placeholders and provides better stage separation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,132 +2,146 @@ language: php
 
 dist: trusty
 
-sudo: required
-
 php:
     - 7.2
 
 jobs:
-  include:
-    - stage: deploy
-      env:
-        DOCKER_COMPOSE_VERSION=1.22.0
-      before_install:
-        - sudo rm /usr/local/bin/docker-compose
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,132 +2,130 @@ language: php
...
+                - bin/console server:run 127.0.0.1:8080 --env=test_cached --docroot=web --router=app/config/router_test_cached.php --no-debug --quiet > /dev/null 2>&1 &
+
+            script:
+                - composer validate --strict
+
+                - bin/phpspec run --no-interaction -f dot
+
+                - bin/console sylius:fixtures:load --no-interaction --env=test_cached --no-debug -vvv
+
+#                - echo "Testing (Behat - brand new, regular scenarios; ~@javascript && ~@todo && ~@cli)" "Sylius"
```
pamil: To be uncommented, need to check the Docker stage first.

PR #251, opened 2018-08-18
Update LICENSE

xElysioN:
Update LICENSE because i think we are in 2018 :trollface:


lchrusciel:
Can you rebase to 1.1? ;)

pamil:
Thanks! 🎉 Guess we should already make it for 2019 so we wouldn't have to remember about it again in a few months 🤣

PR #252, opened 2018-08-20
Bump to Sylius-Standard 3.0-dev

TomasVotruba:
Since https://github.com/Sylius/Sylius-Standard/ already uses YAML, here it can be used too (it's more clear to understand to Symfony community over NEON).

When https://github.com/SyliusLabs/CodingStandard gets tagged, this can be merged with just version change in `composer.json`

### DIFF
--- a/app/TestAppKernel.php
+++ b/app/TestAppKernel.php
@@ -11,7 +11,7 @@
 
 declare(strict_types=1);
 
-require_once __DIR__.'/AppKernel.php';
+require_once __DIR__ . '/AppKernel.php';
 
 use ProxyManager\Proxy\VirtualProxyInterface;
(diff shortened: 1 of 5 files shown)


pamil:
I think that including the dev-master CodingStandard in stable releases is a wrong idea, the bug is fixed in #260 though, thanks for reporting! I'll try to give our coding standard more attention and get it released with Sylius 1.3.

pamil:
There was `@dev` requirement for the coding standard package so that also shall be changed to `^3.0` after the tag.

If it comes to SyliusLabs/CodingStandard, we keep it in different repositories to allow for different release cycles (1.3 for Sylius, 3.0 for CodingStandard). In the future, I'd also like more people to manage those repos.

TomasVotruba:
I see. Thanks for expalining.

PR #253, opened 2018-08-22
Fix Docker image build order

teohhanhui:
The correct build order is important for proper Docker layer caching.


pamil:
Thank you, Teoh! 🎉

Issue #250, opened 2018-08-09
Use Symfony flex and Symfony 4 folder structure


pierre-H:
Hello !

First, thanks for Sylius : it's a great Framework !

When will you update Sylius to Symfony flex and Symfony 4 folder structure ?
The new system is so easy :)

Thanks !

pamil:
Great timing, I started to work on that this morning, see Sylius/Sylius#9643! :)

diimpp:
Hi,

are there any plans for this before 1.3, which is I assume planned for late September? 

I'm starting a new symfony4/sylius project and wondering if I should tackle it myself locally or start with current version and port to flex dir style later on.

diimpp:
Ok, I guess it's not possible to use flex directory structure until Sylius will remove `symfony/symfony` dependency.

teohhanhui:
@diimpp Did you see the linked PR? https://github.com/Sylius/Sylius/pull/9643

diimpp:
@teohhanhui I've did, but Sylius#master still uses `symfony/symfony` https://github.com/Sylius/Sylius/blob/master/composer.json#L60 dependency, which makes it impossible (To my knowledge) to use flex.

I've tried to install Sylius 1.2 and 1.3@dev on `Symfony4-skeleton` without success.

teohhanhui:
Uhh, yes, guess we have to wait for the next PR from @pamil 🥇

pamil:
@diimpp yeah, `symfony/symfony` is getting removed from `sylius/sylius` requirements in a next PR which introduces Flex :)

Bonobomagno:
so you did it on 1.2.6?
Do you think to add documentation of suggested way to manage new projects?

pamil:
The changes will be included in `1.3.0` release, which is scheduled for the very end of September, so there's a plenty of time for polishing and documenting the new flow.

Bonobomagno:
Thanks. this was not clear becouse you are talking about supporting symfony 4 from 1.2. 
You closed the symfony 4 support issue on sylius-standard, making think that the work is done.

I think that will be better to explain that on the readme, something like (still on symfony3)

teohhanhui:
@Bonobomagno Sylius 1.2 supports Symfony 4.1, but it does not use Symfony Flex and is on the old directory structure.

pamil:
Sylius-Standard is on new directory structure and supporting Symfony Flex now.

PR #254, opened 2018-08-25
Using symfony/flex to clear legacy tags, which reduces memory usage

rubenrua:
Memory and CPU time.

See: https://github.com/composer/composer/issues/7577

Before: Memory usage: 397MB (peak: 1569.52MB), time: 31.17s
After: Memory usage: 230.92MB (peak: 263.72MB), time: 5.78s

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
     "require": {
         "php": "^7.2",
         "sylius/sylius": "1.3.*@dev",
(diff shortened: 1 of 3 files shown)


pamil:
Thanks for the PR, introducing Flex was a bit more of work (#250) and this PR needs rebasing, I created #263 to add symfony-require option. 🥇

PR #258, opened 2018-09-18
Switch to Symfony 4 directory structure + add Flex support

pamil:


public/index.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
```
lchrusciel: shouldn't we use strict types in this class too?

pamil: Will solve it in the next PR! 🎉

PR #260, opened 2018-09-19
Fix ECS config file

pamil:


pamil:
It's a bug fix, it does not currently work. Adding it to Travis is a good idea! 👍

PR #263, opened 2018-09-19
Use Flex's symfony-require to improve performance

pamil:
Highly inspired by @rubenrua solution made in #254.


TomasVotruba:
Thanks for sharing, I'll definitely make use of it

PR #265, opened 2018-09-25
[Maintenance] Add extended models for Sylius resources

GSadee:


pamil:
Could you rebase the PR to the 1.3 branch?

JanTvrdik:
* I think you need to override `createTranslation()` method as well
* I think it helps if you split the entities to multiple namespaces, i.e. `App\Entity\Product\ProductVariant` instead of just `App\Entity\ProductVariant`
* It helps greatly if entities do have `@method` annotations with overriden return types to help PhpStan and PhpStorm.

For example, this is how extended `OrderItem` looks in our skeleton.

~~~php
<?php declare(strict_types = 1);

namespace App\Entity\Order;

use App\Entity\Product\Product;
use App\Entity\Product\ProductVariant;
use Doctrine\Common\Collections\Collection;
use Doctrine\ORM\Mapping as ORM;
use Sylius\Component\Core\Model\OrderItem as SyliusOrderItem;


/**
 * @ORM\Entity
 * @ORM\Table(name="sylius_order_item")

 * @method ProductVariant|null        getVariant()
 * @method Product|null               getProduct()
 * @method Order|null                 getOrder()
 * @method Collection|OrderItemUnit[] getUnits()
 * @method Collection|Adjustment[]    getAdjustments(?string $type = null)
 * @method Collection|Adjustment[]    getAdjustmentsRecursively(?string $type = null)
 */
class OrderItem extends SyliusOrderItem
{
}
~~~

pamil:
@GSadee could you find some time to make the build passing?

lchrusciel:
Thank you, Grzegorz! :tada:

blankcheck:
Hi @GSadee , is there any reason for extending models with classes which mostly do nothing new? 
Only thing I can think off is it makes it a bit simpler for developers who need to extend models, but it also introduces issues for ones who want to use xml/yml mapping...
Also, having bunch of files for no good reason other than making it a bit more simpler to extend doesn't sound like a good practice...

Am I missing something here...

thx

PR #269, opened 2018-10-01
Add config for Doctrine mappings

pamil:


config/packages/doctrine.yaml
```diff
@@ -12,3 +12,14 @@ doctrine:
         charset: UTF8
 
         url: '%env(resolve:DATABASE_URL)%'
+
+    orm:
```
pamil: Taken from `doctrine-bundle` recipe.

PR #268, opened 2018-09-27
Docker nginx config fix for symfony4

doctorx32:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #267 
| License         | MIT

PR #271, opened 2018-10-04
Fix Docker setup after moving to new Symfony / Flex directory structure

teohhanhui:
Replaces #268

/cc @doctorx32


public/.htaccess
```diff
@@ -1,25 +1,68 @@
-DirectoryIndex app.php
+# Use the front controller as index file. It serves as a fallback solution when
```
teohhanhui: Should we remove this file? It is available with `symfony/apache-pack` + `symfony/recipes-contrib`: https://github.com/symfony/recipes-contrib/blob/master/symfony/apache-pack/1.0/public/.htaccess

doctorx32: I agree

pamil:
Thanks a lot, Teoh! 🎉

PR #275, opened 2018-10-06
Copy only specifically what we need in Dockerfile

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.3 (or master?)
| Bug fix?        | no
| New feature?    | no?
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,12 +76,18 @@ WORKDIR /srv/sylius
 ARG APP_ENV=prod
 
 # prevent the reinstallation of vendors at every changes in the source code
-COPY composer.json composer.lock ./
+COPY composer.json composer.lock symfony.lock ./
 RUN set -eux; \
 	composer install --prefer-dist --no-autoloader --no-scripts --no-progress --no-suggest; \
 	composer clear-cache
 
-COPY . ./
+# copy only specifically what we need
+COPY bin bin/
+COPY config config/
+COPY public public/
+COPY src src/
+COPY templates templates/
+COPY translations translations/
 
 RUN set -eux; \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -76,12 +76,18 @@ WORKDIR /srv/sylius
...
 
 # prevent the reinstallation of vendors at every changes in the source code
-COPY composer.json composer.lock ./
+COPY composer.json composer.lock symfony.lock ./
 RUN set -eux; \
 	composer install --prefer-dist --no-autoloader --no-scripts --no-progress --no-suggest; \
 	composer clear-cache
 
-COPY . ./
+# copy only specifically what we need
```
teohhanhui: We do not copy directories like `features` and `tests`, because they're not used in production. But they're available in development anyway, since we bind-mount the project directory from host.

Dockerfile
```diff
@@ -76,12 +76,18 @@ WORKDIR /srv/sylius
...
+COPY bin bin/
+COPY config config/
+COPY public public/
+COPY src src/
+COPY templates templates/
+COPY translations translations/
```
adrienlucas: Why not `features` ?

teohhanhui: See comment above: https://github.com/Sylius/Sylius-Standard/pull/275#discussion_r223177966

teohhanhui:
The build failure will be fixed once #276 is merged, because `APP_ENV` would be set.

lchrusciel:
Thanks Teoh!

PR #277, opened 2018-10-06
Only change php memory_limit in dev for Docker

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.3 (or master?)
| Bug fix?        | no
| New feature?    | no?
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

Because memory usage should be under control in `prod`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,7 @@ services:
       - PHP_DATE_TIMEZONE=${PHP_DATE_TIMEZONE:-UTC}
     volumes:
       - .:/srv/sylius:rw,cached
+      - ./docker/php/php-cli.dev.ini:/usr/local/etc/php/php-cli.ini
       # if you develop on Linux, you may use a bind-mounted host directory instead
       # - ./var:/srv/sylius/var:rw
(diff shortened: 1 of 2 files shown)


teohhanhui:
Oh, but we do need to check how much memory is used when doing `bin/console cache:clear` in `prod`.

teohhanhui:
```
# export APP_ENV=prod; /usr/bin/time -f '%M mmem' composer run-script post-install-cmd
ocramius/package-versions:  Generating version class...
ocramius/package-versions: ...done generating version class
Executing script cache:clear [OK]
Executing script assets:install public [OK]
Executing script security-checker security:check [OK]

38756 mmem
```

That's ~38 MiB, so I think we're fine.

pamil:
@teohhanhui have you checked the issue you mentioned in the last comment?

teohhanhui:
I'm not sure how much memory it actually uses, but I can confirm that it's successful with a 128M memory limit. (Not with `prod` environment though as that's currently broken.)

pamil:
Thanks, Teoh! 🎉

PR #281, opened 2018-10-10
Do not include nelmio_alice config into prod env

vvasiloi:


vvasiloi:
Not sure whether addding this file is necessary at all, since it has only a default config in it:
https://github.com/nelmio/alice/blob/16959632cd73c019e2ae4d3ccd88062056738cb7/src/Bridge/Symfony/DependencyInjection/Configuration.php#L59

lchrusciel:
We may try to remove the config and see if it works. Nevertheless, I suppose it is an autogenerated file so that it may stay in a test, dev, and test_cached.

vvasiloi:
I think that this commit explains everything.
https://github.com/symfony/recipes-contrib/commit/03824aa669409cc450d868392515e0ebcc4defd4#diff-f803db5bab8934a062863e026cd6225c

pamil:
Thank you, Victor! 🎉

Issue #296, opened 2018-11-13
sensiolabs/security-checker should be in require packages not in require-dev


Prometee:
I was merging the last changes with my current project but I got stuck while deploying to my staging env.

The composer `auto-scripts` will run `security-checker` command but if you install your dependencies with `composer install --no-dev` the package `sensiolabs/security-checker` won't be installed because it is in the `require-dev`.

PR #302, opened 2018-11-29
Wrong dev package placement

Prometee:
`sensiolabs/security-checker` is needed for `auto-scripts` so when you put it in production you will get errors because this package does not exists.

Bug #296

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     "require": {
         "php": "^7.2",
         "sylius/sylius": "~1.3.0",
-        "symfony/flex": "^1.1"
+        "symfony/flex": "^1.1",
+        "sensiolabs/security-checker": "^5.0"
     },
     "require-dev": {
         "behat/behat": "^3.4",
@@ -38,7 +39,6 @@
         "lakion/mink-debug-extension": "^1.2.3",
         "phpspec/phpspec": "^5.0",
         "phpunit/phpunit": "^6.5",
-        "sensiolabs/security-checker": "^5.0",
         "stripe/stripe-php": "^4.1",
         "sylius-labs/coding-standard": "^2.0",
         "symfony/browser-kit": "^3.4|^4.1",


lchrusciel:
Thanks, Francis! :tada:

teohhanhui:
This was the wrong fix. See https://github.com/Sylius/Sylius-Standard/pull/282

Prometee:
In this fix #282 @pamil add `security-checker` to travis and test env. But test env has to load `require` and `require-dev` packages that why there is no bug at the first time in November. But if you deploy your app with no dev package there will be a problem because `auto-scripts` need this package to be in the `require` not the `require-dev`.

@teohhanhui i Je peux te donner ma configuration capistrano si tu veux tester sur un espace de prod.

teohhanhui:
@Prometee No, he forgot to remove the script from `@auto-scripts` :laughing:

Prometee:
I think we have to keep it. This little script will only check if a package has security risk spotted. If I deploy an app I want to know if there is a problem with any of my dependencies. Sylius take care about there dependencies doing this in the test env. But devs has to be warn when they use some other dependencies.

teohhanhui:
It's not right to run a security check every time you do a `composer install`, especially in `prod`.

Prometee:
Yes you will have to do it while requiring or updating a packag. But what if you merge to your master and this new commit trigger a deployment. Maybe you want to be warn before deploying and maybe not all devs will build a pre-deployment test script.

teohhanhui:
Of course, nothing stops you from doing that in your `prod` environment if you think it's appropriate, but I'd say it should not be what Sylius-Standard does out of the box.

teohhanhui:
Usually you should do as little as possible in `prod` environment. This kind of automated checks belong in your CI / CD.

Prometee:
@pamil maybe Sylius has to make a choice here : keep this package and `auto-script` or remove it from main `composer.json` and keep it in the `test` env.

PR #311, opened 2019-01-04
[Composer] Remove stripe dependency

lchrusciel:


pamil:
Thanks, Łukasz! :tada:

PR #314, opened 2019-01-07
[Composer] Update stripe dependency

lchrusciel:
#314 was missing an update of `composer.lock`. Sorry 💃


pamil:
I think it's a bit outdated now, please reopen if it still applies.

PR #321, opened 2019-01-22
Fix the build vol.2

Zales0123:
Replaces https://github.com/Sylius/Sylius-Standard/pull/320

I've installed again `stripe/stripe-php` which was removed in https://github.com/Sylius/Sylius-Standard/pull/311... but the build was failing, I don't know why 🐰


pamil:
Thanks, Mateusz! :tada:

PR #318, opened 2019-01-19
Fix doctrine migrations

lemuelroberto:
Fix #317 issue.

Context:

The commit 060d1629b3c9261edaf9f48b50da3698008fc605 updated the composer
dependencies and introduced some breakages.

The composer.lock was using the doctrine/doctrine-migrations-bundle:v1.3.2
and the doctrine/migrations:v1.8.1 versions.

The doctrine/migrations package major upgrade to the v2.0.0 version changed
the autoload namespace of AbstractMigration class and started to use the new
PHP 7.0 return type declarations feature.

PR #322, opened 2019-01-22
Update to Sylius 1.4-BETA

Zales0123:
It contains changes from https://github.com/Sylius/Sylius-Standard/pull/318


behat.yml.dist
```diff
@@ -3,41 +3,9 @@ imports:
 
 default:
     extensions:
-        FriendsOfBehat\ContextServiceExtension:
-            imports:
-                - "vendor/sylius/sylius/src/Sylius/Behat/Resources/config/services.xml"
-
+        Behat\MinkExtension:
```
pamil: Could we copy Sylius/Sylius's `behat.yml.dist` contents here? So that we won't import `vendor/sylius/sylius/behat.yml.dist` here but only the suites definitions.

composer.json
```diff
@@ -27,12 +27,9 @@
...
         "behat/mink-extension": "^2.2",
         "behat/mink-selenium2-driver": "^1.3",
-        "friends-of-behat/context-service-extension": "^1.2",
-        "friends-of-behat/cross-container-extension": "^1.1",
-        "friends-of-behat/page-object-extension": "^0.2.1",
-        "friends-of-behat/service-container-extension": "^1.0",
+        "friends-of-behat/page-object-extension": "^0.3",
         "friends-of-behat/suite-settings-extension": "^1.0",
-        "friends-of-behat/symfony-extension": "^1.2.1",
+        "friends-of-behat/symfony-extension": "^2.0.0-rc.2",
```
pamil: It's stable now! 🎉

config/services_test.yaml
```diff
@@ -0,0 +1,3 @@
+
```
pamil: 😱

PR #329, opened 2019-01-28
set sql_mode to empty string in docker composer setup

lsmith77:
work around for https://github.com/Sylius/Sylius/issues/7389

ie. without it using the standard fixtures on http://localhost/en_US/taxons/mugs I get:
```
SQLSTATE[HY000]: General error: 3065 Expression #1 of ORDER BY clause is not in SELECT list, references column 'sylius.s2_.position' which is not in SELECT list; this is incompatible with DISTINCT
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,7 @@ services:
       - MYSQL_DATABASE=sylius
       - MYSQL_USER=sylius
       - MYSQL_PASSWORD=${MYSQL_PASSWORD:-nopassword}
+    command: mysqld --sql_mode=""
     volumes:
       - mysql-data:/var/lib/mysql:rw
       # you may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!


4c0n:
@lsmith77 The link you included to the Issue does not work, could you update the description with the actual url to the issue, please?

lsmith77:
fixed the link and added url and error

4c0n:
By default the `SQL_MODE` is set to `ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION` in percona/mysql 5.7
I think instead of overriding the command used in the docker-compose file it would be better to provide a `my.cnf` file where you set the `SQL_MODE` as above, but without `ONLY_FULL_GROUP_BY` (as that is what is causing the error).

To see how you can do that have a look at the docker hub page for the percona image: https://hub.docker.com/_/percona (section "Using a custom MySQL configuration file").

In the end someone should step up and make sure a query is produced that works with strict mode enabled, which might not happen for a long time if a workaround gets merged (as nobody is experiencing any problems anymore, there would be no desire to fix it).

Just my 2 cents

JanTvrdik:
This must be fixed in Sylius. Changing the SQL mode just hides the error.

lsmith77:
agreed. but the issue was first reported 2 years ago. also this quick fix here is needed in part due to the specific mysql version picked in the docker compose config.

lsmith77:
this issue was fixed .. not sure if there are more strict issue left .. but closing for now

PR #334, opened 2019-03-07
[Docker] Set the mailer to use the mailhog container

spipnl:
Mailhog was not working out of the box for me.
This fixes it. (Or is there a better way to enable mailhog for development?)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
       - APP_DEBUG=1
       - APP_SECRET=EDITME
       - DATABASE_URL=mysql://sylius:nopassword@mysql/sylius
-      - MAILER_URL=smtp://localhost
+      - MAILER_URL=smtp://mailhog:1025
       - PHP_DATE_TIMEZONE=${PHP_DATE_TIMEZONE:-UTC}
     volumes:
       - .:/srv/sylius:rw,cached
@@ -92,6 +92,7 @@ services:
       - ./docker/mailhog/maildir:/maildir:rw,delegated
     ports:
       - "8025:8025"
+      - "1025:1025"
 
 volumes:
   mysql-data:


gabiudrescu:
@CoderMaggie who is in charge of this repo? this PR, imho, should be reviewed and merged.

docker-compose.yml
```diff
@@ -92,6 +92,7 @@ services:
       - ./docker/mailhog/maildir:/maildir:rw,delegated
     ports:
       - "8025:8025"
+      - "1025:1025"
```
teohhanhui: This line should be removed.

spipnl: Could you please explain why? I've added it to made the mailhog service visible for the php service. Hence the other modified line (https://github.com/Sylius/Sylius-Standard/pull/334/commits/be9d23456f0a607f304edc9839fd161c04e22fb0#diff-4e5e90c6228fd48698d074241c2ba760R20).

teohhanhui: That is not how it works. :smile:

teohhanhui: Since the 2 containers are connected to the same network, they can see each other just fine. That's how the `nginx` container connects to the `php` container, for example.

teohhanhui: What you've actually done here with this line is to publish port 1025 from the `php` container to port 1025 on the host, which of course makes no sense.

pamil:
The base of this pull-request was changed, you need fetch and reset your local branch 
if you want to add new commits to this pull request. **Reset before you pull, else commits
may become messed-up.**

Unless you added new commits (to this branch) locally that you did not push yet, 
execute `git fetch origin && git reset "master"` to update your local branch.
 
Feel free to ask for assistance when you get stuck :+1:

pamil:
Thanks, Jip! :1st_place_medal:

pamil:
And thanks Teoh for the review and you all for the patience :)

PR #342, opened 2019-05-09
Synchronize security.yaml with Sylius/Sylius

Zales0123:
The same should be done for `1.4`, I believe it can be a problem for not being able to log in to the administration panel just after the installation (but I've checked it on `1.4`, so I don't know does it happen also for `1.3`):

Screenshot 2019-05-09 at 11 40 10


bartoszpietrzak1994:
Thanks, Mateusz! :1st_place_medal:

PR #346, opened 2019-05-23
[Maintenance] Update php, nginx and apcu versions in Dockerfile

teohhanhui:


Dockerfile
```diff
@@ -1,9 +1,9 @@
 # the different stages of this Dockerfile are meant to be built into separate images
 # https://docs.docker.com/compose/compose-file/#target
 
-ARG PHP_VERSION=7.2
+ARG PHP_VERSION=7.3
 ARG NODE_VERSION=10
```
teohhanhui: node version not upgraded as it's an active LTS: https://github.com/nodejs/Release#release-schedule

Active support is until April 2020. So I think we're fine for now? Or do we want to update to the current release (12)?

pamil: If there aren't any problems connected with upgrading then I'd vote for upgrading. I'll merge this PR in this form, could you open another one with Node upgrade?

pamil:
Thank you, Teoh! :tada:

Dockerfile
teohhanhui: The one that usually gives problems is `node-sass` :laughing:

But yeah, if everything works...

PR #350, opened 2019-06-13
Fix deprecation notice

loevgaard:
Fixes this notice: `The default value of "$usePutenv" argument of "Symfony\Component\Dotenv\Dotenv::__construct" will be changed from "true" to "false" in Symfony 5.0. You should define its value explicitly.`


pamil:
Thank you, Joachim! :1st_place_medal:

PR #358, opened 2019-07-15
Don't mount mailhog maildir directory from host

teohhanhui:
Avoid problem with permissions as the mailhog container does not run as root


pamil:
Thanks, Teoh! :tada:

PR #360, opened 2019-07-15
[Config] Add missing media image folder

lchrusciel:
On fresh copy, there are no ability to list commands and run some other commands until sylius:install will be executed (which will create public/media/image directory)

Ref. https://github.com/Sylius/Sylius/pull/9961


pamil:
Thank you, Łukasz! :1st_place_medal:

PR #369, opened 2019-09-11
Colorize inform about GUS command

Zales0123:
Before:

Zrzut ekranu 2019-09-11 o 09 03 17

After:

Zrzut ekranu 2019-09-11 o 09 03 23


lchrusciel:
Thank you, Mateusz! :1st_place_medal:

PR #372, opened 2019-10-14
Fix custom translations

loic425:


loic425:
@pamil @lchrusciel 
Without this configuration, we can't add custom translations.

pamil:
Thanks, Loïc! :tada:

PR #378, opened 2019-12-05
Fix the Behat build

pamil:


GSadee:
Thank you, Kamil! :tada:

PR #382, opened 2020-02-12
Set scan_depth=1 to avoid going down into the themes directories

jacquesbh:
I gain 14% performance.

image

This value is the default one in Sylius. But I don't know why it wasn't put back.
So, without it the `scan_depth` is equal to `null`, which leads to a lot of files scanned, especially if you have a `node_modules/` directory:

image

These metrics are made with Blackfire on one of our projects. It may differ for others but I think it's a good thing to keep `1` as default in Sylius-Standard.


pamil:
The base of this pull-request was changed, you need fetch and reset your local branch 
if you want to add new commits to this pull request. **Reset before you pull, else commits
may become messed-up.**

Unless you added new commits (to this branch) locally that you did not push yet, 
execute `git fetch origin && git reset "patch-1"` to update your local branch.
 
Feel free to ask for assistance when you get stuck :+1:

pamil:
Thanks, Jacques! :tada:

PR #384, opened 2020-02-20
Update platform php version

alexander-schranz:
Its bad to use here 7.2 or 7.3 as this will force the php version to 7.2.0 ([reference](https://github.com/composer/composer/issues/8144)). For example if a package requires in a newer version `^7.2.1` you will not get the newest version of it, so if make usage of the config.platform.php it should use the 3 digit version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -78,7 +78,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.3.14"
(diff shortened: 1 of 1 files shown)


pamil:
Thanks for the PR! However, it looks like it was fixed in the meantime somehow.

PR #395, opened 2020-04-20
Remove existing queries in last migration

pocky:
Last migration already contains a lot a previous queries.

- Line 25 to 27 are in `Version20200122082429.php`
- Line 28 is in `Version20200202104152.php`
- Line 29 to 31 are in `Version20200110132702.php` (except they don't drop `type`)


lchrusciel:
Hey Alexandre,

thanks a lot for your proposal. As we prefer to not change existing migration, we've decided to introduce a new one. Your problem was fixed in #406. I hope you don't mind and see you soon on another PR/issue ;)

PR #409, opened 2020-06-08
[Travis] Switch to Bionic & fix the build

pamil:


.travis.yml
```diff
@@ -146,14 +118,3 @@ jobs:
                 - docker-compose exec php bin/console sylius:fixtures:load --no-interaction
 
                 - curl http://localhost/
-
-            before_deploy:
```
pamil, line 150: It has never been deployed before and I don't think we can find the time (if anyone would like to give it a try, let me know).

PR #416, opened 2020-07-24
Update composer.lock

Tomanhez:


GSadee:
Thank you, Tomasz! :tada:

PR #434, opened 2020-10-23
[Maintenance] Switch to composer v2

lchrusciel:
This PR aims to reduce execution time and memory usage

PR #436, opened 2020-10-27
[Maintenance] Remove php platform from composer

lchrusciel:
I've removed the platform from the composer to let flex detect the current environment based on used CLI PHP. In addition, I've added composer v2

Replacecment for #434

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ jobs:
 
                 - echo "memory_limit=4096M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
-                - composer self-update --1
+                - composer self-update --preview
 
             install:
                 - if [ ! -z "${SYMFONY_VERSION}" ]; then bin/require-symfony-version composer.json "${SYMFONY_VERSION}"; fi
--- a/src/Migrations/Version20200916093101.php
+++ b/src/Migrations/Version20200916093101.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
                 - echo "memory_limit=4096M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
-                - composer self-update --1
+                - composer self-update --preview
```
Zales0123, line 42: `composer self-update --2`?

src/Migrations/Version20200916093101.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20200916093101 extends AbstractMigration
```
Zales0123, line 10: Shoudn't it be loaded automatically from the `CoreBundle`?

.travis.yml
lchrusciel: I kept preview to have the newest versions, sometimes not stable even. We had it before v2 was announced, but I don't have a strong opinion about it. But if we will change it, I would go with `--stable`

src/Migrations/Version20200916093101.php
lchrusciel: It is opened to 1.7, where we don't have new migrations yet. During upmerge it should be removed

Zales0123: Indeed 👍

Zales0123 (approved):

Zales0123:
Thank you, Łukasz! :1st_place_medal:

PR #433, opened 2020-10-15
enable Xdebug with docker build stage and dc override

lwillems:
Hello

To enable Xdebug just add a new stage to DockerFile and load override.
Original credits go to @dunglas with [docker sf5 repo ](https://github.com/dunglas/symfony-docker)

Hope it could be useful.

Regards

PR #442, opened 2020-11-10
[Docker] Fix build and xdebug support added

lchrusciel:
Continuation of #433


lchrusciel:
Hey @lwillems, it is a little bit better with the docker build, but still it is red. Not I have some issues with database

lwillems:
> Hey @lwillems, it is a little bit better with the docker build, but still it is red. Not I have some issues with database

Hi @lchrusciel you have ideas to fix that ?

lchrusciel:
Not really. It seems, that we are missing doctrine migration step

PR #476, opened 2020-12-23
Prevents click jacking

LucaGallinari:
Actually, there is no protection against click-jacking. Given the nature of the platform (credentials, credit cards informations ecc) i think that this should be the default behavior.
What do you think?

Doc reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -66,3 +66,8 @@ DirectoryIndex index.php
         # RedirectTemp cannot be used instead
     </IfModule>
 </IfModule>
+
+<IfModule mod_headers.c>
+    # Prevent clickjacking
+    Header set X-Frame-Options SAMEORIGIN
+</IfModule>


FDiskas:
I love this. But this is not good for everyone. There is for example dedicated product review page and this can be embedded for some advertisements or landing pages. This will prevent that.

LucaGallinari:
I know @FDiskas , but i think Sylius should (i don't know if they always already do that) follow the security by default approach.

lchrusciel (approved):
Hey Luca, 

your opinion + OWASP recommendation are more than enough for me to be for merging of this PR (ref. https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html)

Zales0123 (approved):

pamil (approved):

pamil:
Thanks, Luca! :tada:

PR #505, opened 2021-02-04
Remove composer.lock from .gitignore

### LINKED ISSUE #503: composer.lock should be removed from .gitignore
Hi all,

I would strongly advise to remove composer.lock from .gitignore

pamil:
Fixes #503.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@
 /docker/mysql/data/
 
 ###> symfony/framework-bundle ###
-/composer.lock
 /.env.*.local
 /.env.local
 /.env.local.php


lchrusciel:
I would like to challenge one thing. The problem that I have with the existence of the `composer.lock` in the repo is that it will be used at the end application to build an app and may be used to build dependencies in CI. Also, this composer.lock may differ depending on env, PHP version, etc. Which versions should be used in the repo? Lowest? How will we ensure that? I'm afraid, we may provide stale dependencies to the end-users. Perhaps we should add this line to the `.gitignore` as a step of Sylius installation (even onliner in documentation seems to be a better solution, or preferably script of composer/sylius install)

pamil:
@lchrusciel the current issue with `composer.lock` being ignored is that new projects aren't set up to commit `composer.lock` in them (so that they have to remove that line from `.gitignore`). We shouldn't commit any `composer.lock` to this repository to prevent the issues you've mentioned.

I'll add a step in GitHub Actions to fail if `composer.lock` is committed to the repository.

blankcheck:
Hi guys, maybe I am missing something,

BUT, having composer.lock in .gitignore (for sylius-standard) actually means that after installation, we have to verify that everything works and remove it from .gitignore.
Verifying can be simple, but also it can be complex if we want to check every aspect of application - why not share that between other users of sylius - if we all have same versions than we have same potential errors.

IMHO, having exact same versions of everything for official releases is a MUST since nobody can guaranty sem-ver rules being respected across so many plugins and libraries being used...

Better stale than sorry...

PR #536, opened 2021-04-01
Dockerfile and docker-compose update for php version 7.4

arti0090:


Dockerfile
```diff
@@ -76,9 +75,9 @@ WORKDIR /srv/sylius
 ARG APP_ENV=prod
 
 # prevent the reinstallation of vendors at every changes in the source code
-COPY composer.json composer.lock symfony.lock ./
```
lchrusciel, line 79: are you sure this should be removed?

arti0090: The composer.lock file is not needed anymore (since composer v2 as i remember?). The composer is creating it when building containers and modify directory on your machine (the composer.lock is created). I might be wrong with my thinking and the composer.lock file might be needed to prevent reinstalation.
When I checked manualy I made some changes in code of application, and run `docker-compose build` and the composer was running from cache ( no reinstallation of vendors happened).

arti0090: If somebody with docker experience is looking here please give an opinion, I am still learning docker 😄

arti0090:
I am looking for some suggestions for this PR:
1. As I find out the quay.io is now private repositories thats why I deleted it. Is this an OK thing?
2. I have changed some lines for nginx to resolve automatically the IP's of containers. Still in my opinion there is an issue that (in my case) I need to wait for about 10minutes for nginx to work properly (otherwise I have 502 bad gateway error). Any solutions?

Any other suggestions are really appreciated 😄

Dockerfile
olimination: According to the composer docs here: https://getcomposer.org/doc/01-basic-usage.md#installing-with-composer-lock it is a good practice to commit your project's composer.lock file to version control. Also it makes sense to copy this file to Docker's context during image build, because composer then don't need to resolve the latest versions of a dependency but simply grabs the locked version of a dependency in the composer.lock file. This also ensures that you have a controlled Docker image build. Else it may happen that some dependencies changes during your Docker image build process which results in another behaviour.

So with the composer.lock file you can have controlled way of updating your dependencies, without it each `composer install` execution may install some other versions of a dependency which may produce some error in your application.

As far as I see this Sylius-Standard repo also does not include such a composer.lock file, so you need to make sure that you have run composer install (or composer create-project...) before you build a Docker image, else the Docker image build will fail with the message that "the composer.lock file does not exist...".

cristi-contiu: Just a small question here, how would you run `composer install` if you rely on the PHP & Composer provided by the Docker image (which, as you said, will fail to built on first run b/c there is no `composer.lock`)?

olimination: I would run it during local development, test the application and commit it to the repository. The CI/CD pipeline then simply uses the lock file for installing the dependencies, executing all the needed tests and finally builds the Docker image. It actually also depends how you would like to use the Docker container. I would strictly make here a separation between a "local development flow with Docker" and the Docker image build for a Production environment. The Docker image build for the Production environment should actually IMHO be built based on the lock file, else you can't be sure which dependency version finally get installed for your application. Also you should use the same PHP and composer version throughout the whole process in dev and Prod build.

If you would use the Docker image only or also for your local development flow then you really don't need to copy the lock file because then it is ok to simply have the composer.json and execute the composer install. But also there you should then commit the generated lock file to the repository to lock down/snapshot your current dependency version graph which works with your current application version.

docker-compose.yml
```diff
@@ -46,11 +42,7 @@ services:
     build:
       context: .
       target: sylius_nodejs
-      cache_from:
-        - quay.io/sylius/php:latest
-        - quay.io/sylius/nodejs:latest
-        - quay.io/sylius/nginx:latest
-    image: quay.io/sylius/nodejs:latest
+    image: nodejs:latest
```
tonicospinelli, line 45: `nodejs` not is available image, the correct is `node`

tonicospinelli:
@arti0090 using docker I faced this prolonged response too. It's probably around php-fpm, the Nginx wait for it. Because of that you getting a Bad Gateway.

tonicospinelli:
I'm trying to help with these little changes at: https://github.com/arti0090/Sylius-Standard/pull/1

Dockerfile
arti0090: thanks @ilimination for your review comment, unfortunately I don't think that composer.lock file can be left. As in Sylius there are is a matrix of version supported (like php 7.4 or 8.0 and different mysql's) and leaving composer lock might interfere with it. Speaking about splitting prod and dev in two - well te bo honest I struggle with my experience at the moment to make it right 😄 .

cristi-contiu: A fix for being able to build the Docker image before `composer install` (without an `composer.lock` file) was proposed here: https://github.com/Sylius/Sylius-Standard/pull/559

arti0090:
@tonicospinelli I was making some investigations about nginx 502 error. It looks like the php container was working too slow, or being precise - nginx and php were starting at the same time but php was than starting doctrine scripts and also building assets. When it finished the nginx was able to show Sylius page with no trouble.

Dockerfile
cristi-contiu: Including `composer.lock` in the Sylius-Standard repo might indeed interfere with the various PHP versions, but only providing instructions to Docker to COPY your own `composer.lock` file once it's initialized should not interfere.

arti0090: I have added the line from #559 into the PR. Credits to you @tarnawski , hope this won't be a problem to use your code here? 😄

docker-compose.yml
arti0090: I have changed any nodejs to node in files.

Dockerfile
tarnawski: A more proper way will be merge https://github.com/Sylius/Sylius-Standard/pull/559 and then rebase your changes to be aline with the master, but as is a small improvement - take it as yours :wink:

gdarquie:
I can not wait to see this PR published, thanks @arti0090 and all reviewers ;-)

lchrusciel (approved):

docker-compose.yml
```diff
@@ -42,15 +43,16 @@ services:
...
-        - quay.io/sylius/php:latest
-        - quay.io/sylius/nodejs:latest
-        - quay.io/sylius/nginx:latest
-    image: quay.io/sylius/nodejs:latest
+      target: sylius_node
+      # Quay does not work, should be replaced in future with f.e. ghcr.io
+      # cache_from:
+      #   - quay.io/sylius/php:latest
+      #   - quay.io/sylius/nodejs:latest
+      #   - quay.io/sylius/nginx:latest
```
lchrusciel, line 54: Commented config (unless left for reference for users) should be removed - as it is just a dead "code"

lchrusciel (approved):

lchrusciel:
Thank you, @arti0090! :tada:

lchrusciel:
Hey folks! Thanks for this amount of interest in Sylius docker - I hope we will be able to forge a better, more sustainable docker image for Sylius with you. Changes from this PR will be release in about ~2 weeks, until then you will have to use dev-master to take advantage of it.

3kynox:
Hello, could it be simply updated to PHP8, as Sylius 1.10 brings it? Any love to Kubernetes as well could be planned / provided?

tonicospinelli:
@lchrusciel I add some adjusts to increase the development experience over docker. Take a look on #568 when you have time 🤗

vasilvestre:
> Hello, could it be simply updated to PHP8, as Sylius 1.10 brings it? Any love to Kubernetes as well could be planned/provided?

I would like PHP 8 too, maybe we could create two tags: PHP 7.4 and 8.0 ?
So people could pick the version they need

docker/nginx/conf.d/default.conf
```diff
@@ -1,14 +1,17 @@
...
         try_files $uri /index.php$is_args$args;
     }
 
     location ~ ^/index\.php(/|$) {
-        # Comment the next line and uncomment the next to enable dynamic resolution (incompatible with Kubernetes)
-        fastcgi_pass php:9000;
+        resolver 127.0.0.11 valid=10s ipv6=off;
+        set $backendfpm "php:9000";
+        fastcgi_pass $backendfpm;
+        # Comment the next line and uncomment the next to enable dynamic resolution (incompatible with Kubernetes);
```
vasilvestre, line 14: Aren't they inversed ?

vasilvestre:
I would really love to get it.. there's more and more issue about this. Please release the tags

vasilvestre:
> Hey folks! Thanks for this amount of interest in Sylius docker - I hope we will be able to forge a better, more sustainable docker image for Sylius with you. Changes from this PR will be release in about ~2 weeks, until then you will have to use dev-master to take advantage of it.

I would really love to get it.. there's more and more issue about this. Please release the tags

PR #541, opened 2021-04-22
Switch to PHP 7.4 only

pamil:


lchrusciel (approved):

composer.json
```diff
@@ -17,10 +17,11 @@
         }
     ],
     "require": {
-        "php": "^7.3",
+        "php": "^7.4",
         "friendsofsymfony/oauth-server-bundle": ">2.0.0-alpha.0 ^2.0@dev",
         "sylius/paypal-plugin": "^1.1",
         "sylius/sylius": "~1.10.0@dev",
+        "sylius/admin-api-bundle": "^1.10",
```
lchrusciel, line 24: But do we really want to have it preinstalled?

pamil: No, I'll remove it in the next PR (with all the related config). I just wanted the build to pass cause it's currently failing.

PR #560, opened 2021-05-27
Speed up php-fpm container on MacOS

tonicospinelli:
The script setup_native_nfs_docker_osx.sh will activate the NFS enabling
to speed up the mounted storage.

Adding health check for php-fpm container, just running the command
`docker-compose ps` it will show if the container is healthiness.

The HEALTHCHECK instruction tells Docker how to test a container
to check that it is still working. This can detect cases such as a
web server stuck in an infinite loop and unable to handle new
connections, even though the server process is still running.

Separating some commands into RUN instructions allow the Docker to cache
the instruction result to reuse in another moment.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache

Co-authored-by: Kévin Dunglas <dunglas@gmail.com>
Co-authored-by: arti0090 <arti0090@gmail.com>

### DIFF
--- a/.env
+++ b/.env
@@ -12,7 +12,7 @@ APP_SECRET=EDITME
 # Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
 # For a sqlite database, use: "sqlite:///%kernel.project_dir%/var/data.db"
 # Set "serverVersion" to your server version to avoid edge-case exceptions and extra database calls
-DATABASE_URL=mysql://root@127.0.0.1/sylius_%kernel.environment%
+DATABASE_URL=mysql://sylius:nopassword@mysql/sylius
 ###< doctrine/doctrine-bundle ###
 
 ###> symfony/swiftmailer-bundle ###
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,64 +1,70 @@
 # the different stages of this Dockerfile are meant to be built into separate images
+# https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage
 # https://docs.docker.com/compose/compose-file/#target
 
-ARG PHP_VERSION=7.3
+# https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
+ARG PHP_VERSION=7.4
 ARG NODE_VERSION=10
-ARG NGINX_VERSION=1.16
+ARG NGINX_VERSION=1.17
 
+# "php" stage
 FROM php:${PHP_VERSION}-fpm-alpine AS sylius_php
 
 # persistent / runtime deps
 RUN apk add --no-cache \
-		acl \
-		file \
-		gettext \
-		git \
-		mariadb-client \
-	;
+        acl \
+        fcgi \
+        file \
+        gettext \
+        git \
+        mariadb-client \
+    ;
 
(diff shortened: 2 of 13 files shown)


tonicospinelli:
Facing many issues over docker provision, I think MacOS users may use docker containers without so much pain and a slow interface.

I made this branch from 1.9 just because the master is under construction to migrate to PHP 8.

this PR was inspired by a good job made by @arti0090 on #536

Dockerfile
```diff
@@ -101,12 +113,20 @@ VOLUME /srv/sylius/var
 
 VOLUME /srv/sylius/public/media
 
+COPY docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
+RUN chmod +x /usr/local/bin/docker-healthcheck
+
+HEALTHCHECK --interval=10s --timeout=3s --retries=3 CMD ["docker-healthcheck"]
```
tonicospinelli, line 119: it helps to know when php-fpm is ready

.env
```diff
@@ -12,7 +12,7 @@ APP_SECRET=EDITME
 # Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
 # For a sqlite database, use: "sqlite:///%kernel.project_dir%/var/data.db"
 # Set "serverVersion" to your server version to avoid edge-case exceptions and extra database calls
-DATABASE_URL=mysql://root@127.0.0.1/sylius_%kernel.environment%
+DATABASE_URL=mysql://sylius:nopassword@mysql/sylius
```
arti0090, line 15: Hi @tonicospinelli , may I ask why would you want to change it?

docker/php/docker-entrypoint.sh
```diff
@@ -7,16 +7,6 @@ if [ "${1#-}" != "$1" ]; then
...
 if [ "$1" = 'php-fpm' ] || [ "$1" = 'bin/console' ]; then
-	mkdir -p var/cache var/log public/media
-	setfacl -R -m u:www-data:rwX -m u:"$(whoami)":rwX var public/media
-	setfacl -dR -m u:www-data:rwX -m u:"$(whoami)":rwX var public/media
-
-	if [ "$APP_ENV" != 'prod' ]; then
-		composer install --prefer-dist --no-progress --no-suggest --no-interaction
-		bin/console assets:install --no-interaction
-		bin/console sylius:theme:assets:install public --no-interaction
-	fi
```
tonicospinelli, line 18: all these commands already ran during the container build. so, if you need to update some file, just run `docker-composer build php` to update files inside the container

.env
tonicospinelli: I changed it to give a smooth experience for new adopters because the docker is not parameterized for multi-environment yet.

vvasiloi:
@tonicospinelli @arti0090 guys, maybe join forces and split your changes so at least some can be merged sooner.
- In https://github.com/Sylius/Sylius-Standard/pull/536 only leave changes related to PHP 7.4
- In this PR only leave the NFS setup for MacOS
- Other changes separately
Otherwise none of it will be merged soon.

vvasiloi:
And remove `composer.lock` until there is a resolution on this https://github.com/Sylius/Sylius-Standard/pull/505

tonicospinelli:
I'm closing it to keep only changes related to NFS approach, see you on next PR.

PR #567, opened 2021-06-01
Fix Doctrine config

pamil:
It breaks the application when DoctrineBundle v2.4.0+ is used. See https://github.com/doctrine/DoctrineBundle/pull/1356.


GSadee:
Thank you, Kamil! :tada:

PR #568, opened 2021-06-02
Speed up PHP-fpm container  on MacOS

tonicospinelli:
Many MacOS users may be facing slow responses from Shop or Admin when
running the sylius application in the container. The main reason is how
docker was build in the MacOS architecture. The PHP application has
thousands of files to be loaded in memory (vendor directory mainly).

So, to decrease the page response time, the vendor directory is not
synced into the container. It will be ignored by docker-composer.yml
using the named volume.

Also, was added a health check for the PHP-fpm container to know when it
is ready. If you want to check healthiness, just run the command
`docker-compose ps` to see the container State.

The HEALTHCHECK instruction tells Docker how to test a container
to check that it is still working. This can detect cases such as a
web server stuck in an infinite loop and unable to handle new
connections, even though the server process is still running.

The PHP and PHP-FPM configurations received some adjustments to get
better support development environment experience using containers.

I did pick up these updates from two other contributors that I
co-authored in this commit.

Co-authored-by: Kévin Dunglas <dunglas@gmail.com>
Co-authored-by: arti0090 <arti0090@gmail.com>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,21 +5,16 @@ services:
     build:
       context: .
       target: sylius_php
-        # Quay does not work, should be replaced in future with f.e. ghcr.io
-        # cache_from:
-        #   - quay.io/sylius/php:latest
-        #   - quay.io/sylius/nodejs:latest
-        #   - quay.io/sylius/nginx:latest
     image: php:latest
+    healthcheck:
+        interval: 10s
+        timeout: 3s
+        retries: 3
+        start_period: 30s
     depends_on:
       - mysql
-    environment:
-      - APP_ENV=dev
-      - APP_DEBUG=1
-      - APP_SECRET=EDITME
-      - DATABASE_URL=mysql://sylius:nopassword@mysql/sylius
-      - MAILER_URL=smtp://mailhog:1025
-      - PHP_DATE_TIMEZONE=${PHP_DATE_TIMEZONE:-UTC}
+    env_file:
+        - .env
     volumes:
       - .:/srv/sylius:rw,cached
       # if you develop on Linux, you may use a bind-mounted host directory instead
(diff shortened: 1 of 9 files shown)


tonicospinelli:
In the environment with `APP_DEV = dev` and ` APP_DEBUG = 1`, the response time is around ~2s or more.
As for `APP_DEV = prod` and ` APP_DEBUG = 0`, the response time is about ~300ms or more.

Of course, it depends on your workstation and the resources available for the containers. Therefore, I got a better experience in the local environment with these changes.

Always the first request took more time than I mentioned above, it is because the PHP is caching code on opcache and Symfony generating the cache files.

.env.prod
```diff
@@ -1,2 +1,3 @@
+APP_ENV=prod
```
vvasiloi, line 1: This file is only loaded when the env = prod, so it's redundant.

tonicospinelli: @vvasiloi I forgot to remove 🤗

docker-compose.yml
```diff
@@ -28,13 +28,15 @@ services:
...
+      - sylius-node_modules:/srv/sylius/node_modules
 
   mysql:
     image: percona:5.7
     environment:
       - MYSQL_ROOT_PASSWORD=${MYSQL_ROOT_PASSWORD:-nopassword}
-      - MYSQL_DATABASE=sylius
-      - MYSQL_USER=sylius
+      - MYSQL_DATABASE=sylius_${APP_ENV}
+      - MYSQL_USER=${MYSQL_ROOT_PASSWORD:-sylius}
```
vvasiloi, line 39: Unedited copy/paste?

tonicospinelli: Nice catch! my fault to not review it before pull

julien-hoste (approved):

tonicospinelli:
@vvasiloi I fix the typos 😊

victor-v-bv (approved):

vvasiloi (approved):

lchrusciel:
@Arminek can you take a look into this?

Ferror:
It's a nice improvement, but with the latest Docker Desktop filesystem, there is no need for performance improvement.

https://www.docker.com/blog/speed-boost-achievement-unlocked-on-docker-desktop-4-6-for-mac/

I don't mind keeping the health checks in the environment :)

jacquesbh:
@Ferror Docker for mac still has some issues unresolved :). But soonish I hope!

rust-le:

PR #580, opened 2021-06-21
Disable auth for LiipImagineBundle controllers

mbabker:
Ref: https://twitter.com/mbabker/status/1405962888792231939?s=21 and https://github.com/liip/LiipImagineBundle/pull/1383

Most apps don't need authenticated user info when resolving images, so we can sanely disable the security authentication for these routes.


lchrusciel (approved):

lchrusciel:
Thanks, Michael! :tada:

PR #605, opened 2021-08-31
Update php, nodejs and nginx in Dockerfile

4c0n:
Updating the default versions in the Dockerfile.

The composer dependencies listed by Sylius standard no longer support PHP 7.3 and the NodeJS + nginx versions are heavily outdated.

Upgrading NodeJS even further can be done, but requires a lot more work as the front end dependencies are also pretty outdated. So I'd suggest doing it in a separate PR or development cycle.

Please don't just tag this and leave it hanging, currently the Docker setup is unusable because of this.
Perhaps we should add a step to the github actions workflow to check and confirm that the container images can actually be built, or even better build them and use them in the workflow. What do you think about that?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 # the different stages of this Dockerfile are meant to be built into separate images
 # https://docs.docker.com/compose/compose-file/#target
 
-ARG PHP_VERSION=7.3
-ARG NODE_VERSION=10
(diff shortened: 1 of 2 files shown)


arti0090 (approved):

lchrusciel (approved):

lchrusciel:
Thanks, Youri! :1st_place_medal:

PR #616, opened 2021-10-21
Remove deprecated whitespace from tag filter

4c0n:
This removes the deprecation warning about the whitespace when running Behat.


Fabiomad85:
Also in 1.10 the problem persists

Zales0123:
Thank you, @4c0n! 🏅

PR #621, opened 2021-11-14
Update python package name in nodejs step

Dragnucs:
Recent alpine updates require python 2 to be installed as `python3` package, otherwise image build fails.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -117,7 +117,7 @@ RUN set -eux; \
 		gcc \
 		git \
 		make \
-		python \
+		python2 \
 	;
 
 # prevent the reinstallation of vendors at every changes in the source code


Dockerfile
```diff
@@ -117,7 +117,7 @@ RUN set -eux; \
 		gcc \
 		git \
 		make \
-		python \
+		python2 \
```
sad270, line 120: It should be python3 not python2

sad270: I'm sorry `python2` works too, i fix the build by using `python3`

lchrusciel (approved):

lchrusciel:
Thank you, @Dragnucs! :tada:

Issue #614, opened 2021-10-15
composer.lock is missing


nfacciolo:
Sylius version : 1.10

during docker installation at Step `Step 16/31 : COPY composer.json composer.lock symfony.lock ./
`
`Error : COPY failed: file not found in build context or excluded by .dockerignore: stat composer.lock: file does not exist`

And indeed, in the root project, there is no composer.lock.

Ferror:
Fixed

Issue #606, opened 2021-09-03
Installation with Docker failed: composer.lock not found


kniziol:
### Introduction

Cannot install Sylius with Docker as mentioned on the https://sylius.com/download page:

```bash
git clone https://github.com/Sylius/Sylius-Standard.git
cd Sylius-Standard
docker-compose --version
docker-compose pull --ignore-pull-failures || true
docker-compose build --pull
docker-compose up -d
docker-compose exec php bin/console sylius:fixtures:load --no-interaction
open http://localhost/
```

### TL;DR

Command `docker-compose build —pull` failed, because `"/composer.lock" not found`. Details below.

### 1st issue

```bash
$ docker-compose pull --ignore-pull-failures || true
Pulling mysql   ... done
Pulling php     ... error
Pulling nodejs  ... error
Pulling nginx   ... error
Pulling mailhog ... done

ERROR: for php  unauthorized: access to the requested resource is not authorized

ERROR: for nodejs  unauthorized: access to the requested resource is not authorized

ERROR: for nginx  unauthorized: access to the requested resource is not authorized
ERROR: unauthorized: access to the requested resource is not authorized
unauthorized: access to the requested resource is not authorized
unauthorized: access to the requested resource is not authorized
```

### 2nd issue

```bash
$ docker-compose build --pull
mysql uses an image, skipping
mailhog uses an image, skipping
Building php
[+] Building 6.1s (19/30)

# ...

 => CACHED [sylius_php  8/20] WORKDIR /srv/sylius                                                                                                                                               0.0s
 => ERROR [sylius_php  9/20] COPY composer.json composer.lock symfony.lock ./                                                                                                                   0.0s
------
 > importing cache manifest from quay.io/sylius/nodejs:latest:
------
------
 > importing cache manifest from quay.io/sylius/nginx:latest:
------
------
 > [sylius_php  9/20] COPY composer.json composer.lock symfony.lock ./:
------
failed to compute cache key: "/composer.lock" not found: not found
ERROR: Service 'php' failed to build : Build failed
```

### Versions

```bash
$ docker --version
Docker version 20.10.8, build 3967b7d

$ docker-compose --version
docker-compose version 1.29.2, build 5becea4c
```

Tlecomte13:
### Solution 

You have to remove the composer.lock in the .gitignore and add this same file in the branch

kniziol:
[...]

Issue #596, opened 2021-07-23
Can't build docker


moezkorkmaz:
PHP version needed >=7.4
Actual PHP version in container 7.3

christopheblin:
A little more details : changing the version from 7.3 to 7.4 is not enough to have the docker build succeed

    #6 24.15 configure: error: unrecognized options: --with-jpeg-dir, --with-png-dir, --with-webp-dir, --with-freetype-dir

have a look to https://github.com/docker-library/php/issues/912 to solve this 

    docker-php-ext-configure gd --with-jpeg --with-webp --with-freetype

but then another error ...

    #6 25.78 configure: error: unrecognized options: --with-libzip

have a look to https://github.com/laradock/laradock/issues/2453#issue-544213537

    docker-php-ext-configure zip

then, I remove composer.lock since it is not part of the git (it should work only with composer.json anyway)

    COPY composer.json symfony.lock ./

But then it fails on composer install (I removed --no-suggest)

```
=> ERROR [sylius_php 10/20] RUN set -eux;  composer install --prefer-dist --no-autoloader --no-scripts --no-progress;  composer clear-ca  16.9s
------                                                                                                                                           
 > [sylius_php 10/20] RUN set -eux;     composer install --prefer-dist --no-autoloader --no-scripts --no-progress;      composer clear-cache:    
#15 0.329 + composer install --prefer-dist --no-autoloader --no-scripts --no-progress                                                            
#15 0.564 No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.                                                                                                                          
#15 0.565 Loading composer repositories with package information                                                                                 
#15 3.697 Updating dependencies
#15 16.70 Killed
```

I tried

    # https://getcomposer.org/doc/03-cli.md#composer-memory-limit
    ENV COMPOSER_MEMORY_LIMIT=-1

But the same error occurs -> I'm out of ideas ...

christopheblin:
A bit more porgress thanks to https://stackoverflow.com/a/50770267/1545567

Increasing the Docker desktop RAM to 3Go now results in successfull build for sylius_php and sylius_nginx

I'll make a PR

vasilvestre:
Duplicate of https://github.com/Sylius/Sylius-Standard/issues/437 ?

lchrusciel:
Hey folks, can you test #605? If it will solve your issue, I can release a new version of Sylius-Standard

[...]

PR #622, opened 2021-11-14
Fix container build missing composer.lock

Dragnucs:
Do not copy non existent composer.lock. The image build correctly without this file anyways.

Related issues: #596, #606, #614.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,7 +76,7 @@ WORKDIR /srv/sylius
 ARG APP_ENV=prod
 
 # prevent the reinstallation of vendors at every changes in the source code
-COPY composer.json composer.lock symfony.lock ./
+COPY composer.json symfony.lock ./
 RUN set -eux; \
 	composer install --prefer-dist --no-autoloader --no-scripts --no-progress; \
 	composer clear-cache


coldic3:
Hi, I faced up the same problem. I'm not sure we should get rid of this part. The main idea behind copying `composer.lock` is to guarantee consistent package versions. Without it, every single build will cause a potentially unintended silent composer update. Silent because the lock file won't change but the vendor will and it will be very confusing and hard to detect in real projects.

Look at the [Dockerfile in API Platform template](https://github.com/api-platform/api-platform/blob/ac68010c818bde422b97a7044b8df04176e970a4/api/Dockerfile#L84) where's the same `COPY` but the difference is that `composer.lock` is in the repository.

Another way I see is to copy `composer.lock` only if the file exists. Unfortunately, it is possible only by some workarounds like using wildcards. Wildcard character set expression seems to me to be the safest option. What do you think to just replace `composer.lock` with `compose[r].lock` or something like that? It solves our problem without any side effects probably.

lchrusciel (approved):

lchrusciel:
Hey @Dragnucs, thanks a lot for proposal. 

Kevin, thanks for checking it up and making research. Keeping composer.lock is no go for us right now, as we are supporting different stacks and it would require different locks for different versions. We may consider providing lock file for docker only if it can be do effort-free during release. Can you confirm, that this fix the main issue?

lchrusciel:
Oh, this is already fixed on master branch. Can you check it?

vasilvestre:
On master, COPY instruction now use "composer.*" instead of "composer.json composer.lock". I've tried to build master minutes ago and it went just fine.

coldic3:
Yeah, the solution with `composer.*` fixes the main issue :) However, it's worth mentioning it copies all the files starting with "composer." so it's good only as far as `composer.json` and `composer.lock` are the only files with "composer." prefix in the root directory.

Zales0123:
Already fixed in https://github.com/Sylius/Sylius-Standard/pull/625, thank you, everyone, for the discussion and your contribution! 🎉 🚀

eradionov:
Was it really fixed?
Still occurs for me

Issue #630, opened 2021-12-03
Cannot install Sylius from CLI


cibolog:
The commands provided at `README.md` fail at `php composer.phar create-project sylius/sylius-standard project`
Ubuntu 20.04, PHP 7.4

```
PHP CodeSniffer Config installed_paths set to ../../slevomat/coding-standard
phpstan/extension-installer: Extensions installed
> phpstan/phpstan-doctrine: installed
> phpstan/phpstan-webmozart-assert: installed
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!  
!!  In CheckExceptionOnInvalidReferenceBehaviorPass.php line 86:
!!                                                                                 
!!    The service "security.authorization_checker" has a dependency on a non-exis  
!!    tent service "security.authentication_manager". Did you mean one of these:   
!!    "security.authorization_checker", "security.authentication_utils", "securit  
!!    y.authentication.manager", "security.authentication.guard_handler"?          
!!                                                                                 
!!  
!!  
Script @auto-scripts was called via post-update-cmd
```

merbekta:
Hi, did you find a solution for this error? i have the same problem

merbekta:
> 

Did you find a solution for this error? i have the same problem

Zales0123:
Hello @cibolog and @merbekta, I think this PR https://github.com/Sylius/Sylius-Standard/pull/631 will solve this issue 🖖

merbekta:
> Hello @cibolog and @merbekta, I think this PR #631 will solve this issue 🖖

Hey @Zales0123 , thanks. You are my hero :)  I used 5.3 instead of symfony 5.4 and my problem was solved

Zales0123:
Great :) I had a quick discussion with @lchrusciel and (as you can see on the linked PR), we will try to introduce proper Symfony 5.4 support to Sylius 1.10 as well (not only upcoming 1.11), so this will not be an issue in the nearest future anymore 💃 Cheers 🖖

merbekta:
> Great :) I had a quick discussion with @lchrusciel and (as you can see on the linked PR), we will try to introduce proper Symfony 5.4 support to Sylius 1.10 as well (not only upcoming 1.11), so this will not be an issue in the nearest future anymore 💃 Cheers 🖖

thank you for your work :)

PR #631, opened 2021-12-06
Require Symfony 5.3 for Sylius 1.10

Zales0123:
As Symfony 5.4 will be [properly supported in Sylius 1.11](https://github.com/Sylius/Sylius/pull/13339), we need to require Symfony 5.3 for Sylius 1.10. If Symfony 5.4 is used, after creating a project there is an exception about problems with the security bundle (see more [here](https://github.com/Sylius/Sylius/issues/13342))

Fixes https://github.com/Sylius/Sylius-Standard/issues/630

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
             fail-fast: false
             matrix:
                 php: ["7.4", "8.0"]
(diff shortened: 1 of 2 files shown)


Zales0123:
Not neede after https://github.com/Sylius/Sylius/releases/tag/v1.10.6 🎉

PR #633, opened 2021-12-08
Add missing polishsymfonycommunity/symfony-mocker-container dependency

jakubtobiasz:
Sylius 1.10.6 release has broken running the app based on the Sylius-Standard in a test environment. It's caused by [moving `"polishsymfonycommunity/symfony-mocker-container"` to `require-dev`](https://github.com/Sylius/Sylius/commit/673bc6662c3934f3e5c1e8cfa0a3fc8710a437b1#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34). `polishsymfonycommunity/symfony-mocker-container` is no longer being downloaded so it throws error because in test env it is used in `src/Kernel.php`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,8 @@
         "symfony/browser-kit": "^4.4 || ^5.2",
         "symfony/debug-bundle": "^4.4 || ^5.2",
         "symfony/intl": "^4.4 || ^5.2",
-        "symfony/web-profiler-bundle": "^4.4 || ^5.2"
+        "symfony/web-profiler-bundle": "^4.4 || ^5.2",
+        "polishsymfonycommunity/symfony-mocker-container": "^1.0"
     },
     "config": {
         "preferred-install": {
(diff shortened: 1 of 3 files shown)


lchrusciel (commented):
Hey Jakub, thanks a lot for your contribution. 

If I see correctly, the PR is failing due to this class: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ApiBundle/EventSubscriber/KernelRequestEventSubscriber.php

Can you debug it and fix the build? I assume, that following this upgrade: https://github.com/Sylius/Sylius/blob/master/UPGRADE-API-1.10.md#upgrade-from-v1100-to-v1101 should do the trick

jakubtobiasz:
@lchrusciel I've enabled Sylius API in `services_test.yaml`. I think it's the best place because we achieve passed tests along with Sylius API disabled by default in an app.

I've run all tests on my fork and every one passed.

lchrusciel:
Thanks, Jakub! :1st_place_medal:

PR #637, opened 2021-12-21
Require doctrine/dbal:^2.7

Zales0123:
Same as https://github.com/Sylius/Sylius/pull/13252, without it, after running `composer create-project ...` and then applying migrations, there is a nasty exception:

Zrzut ekranu 2021-12-21 o 14 36 27

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
     ],
     "require": {
         "php": "^7.4 || ^8.0",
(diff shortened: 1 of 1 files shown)


lchrusciel (approved):

Zales0123:
In fact... it's not needed 😄 when we resolve problems with Sf 6 conflicts, 1.10.7 would be installed and it would be working automatically 🖖

PR #643, opened 2021-12-30
[Docker] Remove python dependency

lchrusciel:


Zales0123 (approved):

loic425 (approved):

Zales0123:
Thanks, Łukasz! :tada:

laurentdabbb:
Hi,

Are you sure you can remove Python ?
`node-gyp@^3.8.0` is in your yarn.lock and node-gyp require Python...

I'm not an expert in Node at all, but I'm having a problem with Python and node-gyp with my customized Dockerfile. I tried to remove python dependency like you do and I encountered an error on build because Python is missing...so I prefer ask for confirmation ;-)

Thx ;-)

PR #659, opened 2022-01-19
Upgrade required version of PHP to 8.0

rimas-kudelis:
~~This updates composer.json to match the requirements in Sylius/Sylius master: PHP `^8.0` and Symfony `^5.4`.~~

~~Also adds a the `extra/symfony/require` element with the value of `^5.4`. This will mean that whenever `composer require symfony/whatev` is issued, it will automatically require `^5.4` when possible, so you won't end up with a mix of packages from 5.4 and 6.x.~~

~~Also bumps `symfony/flex` requirement to `^2.0` (fixes #645).~~

This updates composer.json to require PHP ^8.0.

Also changes `dev-master` branch alias to `1.11-dev` (fixes #635). Although maybe that should already be changed to `1.12-dev` instead?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         }
     ],
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.0",
         "sylius/paypal-plugin": "^1.2.1",
         "sylius/sylius": "^1.11@dev",
         "symfony/dotenv": "^4.4 || ^5.2",
@@ -69,7 +69,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.10-dev"
+            "dev-master": "1.11-dev"
         },
         "symfony": {
             "allow-contrib": false


loic425 (approved):

rimas-kudelis:
@loic425 I forgot to remove 4.4 from the build workflow . I guess that's why these builds were failing.

Roshyo (approved):

.github/workflows/build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
             fail-fast: false
             matrix:
                 php: ["8.0"]
-                symfony: ["^4.4", "^5.4"]
+                symfony: ["^5.4"]
```
lchrusciel, line 27: This change should be done after branching v1.11. Please revert

lchrusciel (commented):
Hey Rimas,

thanks a lot for your contribution! However, I need you to split it into two PR. The first one, should contain only PHP & branch alias change in composer.json. This will land in master, then we will create a new branch - 1.11. After that, we can merge second PR with drop of Symfony 4.4 support - with 1.11 it can still be installed, so we can only do it on master (from 1.12 :))

.github/workflows/build.yml
rimas-kudelis: @lchrusciel Hmmmmmm, so you're suggesting to keep same requirements in Standard and Sylius?

I thought Standard is meant to be a fork-and-forget template for new projects, and thus there was no reason to keep same "relaxed" requirements in it as in sylius/sylius (which can be used during upgrades from older versions, and so it makes sense to support the older LTS there). Is there a flaw in that way of thinking?

loic425: That makes sense. Indeed, on new project, starting from 4.4 is a no way to go. @lchrusciel WDTY?

lchrusciel: Are you sure, that everyone may start with Sf 5.4 deps and the related rest? For all these years related Sylius-Standard template had the same constraints as the main Sylius package. I'm not insisting on this tho

loic425: @vvasiloi Do you have an opinion for that?

vvasiloi: I agree with Lukasz. Testing Sylius-Standard with Symfony 4 will catch issues the users can have when updating an older project where Sylius was updated, but Symfony wasn't.
Sometimes it's difficult to update both at once because of other dependencies, both own and 3rd party.

rimas-kudelis: That is a fair point. I'll update the PR.

rimas-kudelis:
This will be the first part then.
One thing I'm still not sure of is whether to keep `dev-master` branch alias at `1.11-dev` or `1.12-dev`.

lchrusciel:
Thank you, Rimas! :1st_place_medal:

PR #660, opened 2022-01-19
Migrate security config to the new SF5.3 authenticator-based system

rimas-kudelis:
This patch migrates security config to the new authenticator-based system introduced in Symfony 5.1.
Guard-based authentication has been deprecated in Symfony 5.3. Luckily, not much needs to change for us since Sylius didn't have any custom Guard authenticators to begin with.

The patch also drops `always_authenticate_before_granting: true` from config. This option [has been deprecated](https://symfony.com/doc/5.4/reference/configuration/security.html#always-authenticate-before-granting) since Symfony 5.4, and it's unclear why or whether it was needed in the first place, and it caused me personally some headache (see symfony/symfony#43375).

Requires #659 for the sake of consistency.

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -11,7 +11,7 @@ parameters:
     sylius.security.new_api_user_account_regex: "^%sylius.security.new_api_user_account_route%"
 
 security:
-    always_authenticate_before_granting: true
+    enable_authenticator_manager: true
     providers:
         sylius_admin_user_provider:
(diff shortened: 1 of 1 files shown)


Ferror (approved):

diimpp:
Thank you, @rimas-kudelis and sorry for the wait. :pray: 

Solved since then in https://github.com/Sylius/Sylius-Standard/pull/825

PR #663, opened 2022-01-21
Drop symfony4 support

rimas-kudelis:
This is a spin-off from #659 and includes #659.

It updates composer.json:

1. All Symfony packages which currently require v4.4 or v5.x will now require `^5.4`.
2. The `extra/symfony/require` element is added with the value of `^5.4`. This will mean that whenever `composer require symfony/*` is issued, it will automatically require `^5.4` when possible, so you won't end up with a mix of packages from 5.4 and 6.x.
3. `symfony/flex` requirement bumped to `^2.1` (contains #661 and deprecates #645).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,11 +27,8 @@
         "php": "^8.0",
         "sylius/paypal-plugin": "^1.2.1",
         "sylius/sylius": "^1.11@dev",
-        "symfony/dotenv": "^4.4 || ^5.2",
-        "symfony/flex": "^1.11"
-    },
-    "conflict": {
-        "symfony/form": "4.4.11 || 4.4.12"
+        "symfony/dotenv": "^5.4",
+        "symfony/flex": "^2.1"
     },
     "require-dev": {
         "behat/behat": "^3.7",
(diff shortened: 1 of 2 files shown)


loic425 (approved):

loic425:
Oh I'm surprised symfony require config parameter was missing.

loic425:
Indeed, without it, you can have some package with symfony 6 version

composer.json
```diff
@@ -69,10 +66,11 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.10-dev"
+            "dev-master": "1.11-dev"
```
lchrusciel, line 69:
```suggestion
            "dev-master": "1.12-dev"
```

composer.json
```diff
@@ -69,10 +66,11 @@
...
         },
         "symfony": {
-            "allow-contrib": false
+            "allow-contrib": false,
+            "require": "^5.4"
```
lchrusciel, line 73: I do not have strong opinion about it but I'm concerned about this change. Until now, it is fine, because we have just one version, so it will protect us from Sf6. It is also recommended to have it in end projects. However, once we will start supporting Sf6, what value should be placed here? We still can discuss it later, once it will be supported, but I just wanted to bring this topic up

lchrusciel (commented):
Lets start preparing to the next release ;)

composer.json
rimas-kudelis: I've changed it, but I wonder if this is even necessary at all. Does this alias serve any purpose?

lchrusciel (approved):

composer.json
rimas-kudelis: I don't have a good answer to this one. I'm tempted to keep something here to avoid mixed-version installations, but if that would block GitHub pipelines from even testing with SF6, it wouldn't be good.

That said, Sylius Standard is a template for an end project (this repository is even marked as a template already). As such, I would expect it to follow recommendations for end projects. I would expect people to know enough to be able to change this value in their own downstream repositories, should they want to. And for those who don't know how, or have no preference, I think a sane default provided by us would do just fine (and that sane default would probably be the latest Symfony LTS supported by Sylius).

PR #666, opened 2022-01-25
[Docker] Bring back python2 dep

lchrusciel:
Revert of #643. It is still needed due to node and gyp dependency

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -118,6 +118,7 @@ RUN set -eux; \
 		gcc \
 		git \
 		make \
+		python2 \
 	;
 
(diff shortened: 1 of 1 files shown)


Zales0123:
Thanks, Łukasz! :tada:

Ferror:
@lchrusciel I would kindly recommend creating a pipeline that could build the docker image and make sure that at least it builds properly 😢 I see that this topic comes back like a boomerang without any proper comments.

lchrusciel:
This time I at least mention, that it is required by node dependency ;) Other than that, of course! If you, however, would like to contribute something (may be just review or thoughts) please join our slack and write it on `#docker` channel

PR #682, opened 2022-02-27
Introduce composer-unused check

mmenozzi:


vvasiloi (approved):

lchrusciel:
Hey @mmenozzi - can we exclude flex from the check?

mmenozzi:
Hi @lchrusciel, sorry for my late reply...
`symfony/flex` is already ignored (see [here](https://github.com/Sylius/Sylius-Standard/pull/682/files#diff-2ea92a3e5d5f06bcc2387d8c7343eb22940ef3b6da974d605897924c49c52d9dR12)) indeed the build is not failing. But it appears that the `github` output mode puts a warning annotation on the `symfony/flex` package anyway.

This is the raw output using the `default` mode and the `github` mode:

Schermata 2022-08-04 alle 08 02 01

If you prefer I can switch to the `default` output but we'll loose error annotations in the CI summary, like this:


Schermata 2022-08-04 alle 08 04 35

Maybe it's better to loose errors in CI summary instead of having a warning for symfony/flex for every build...

Or maybe we can do this "hack" to exclude warnings from the `github` output:

```
php composer-unused.phar -o github | grep -v "::warning"
```

Let me know.

Rafikooo:
`composer-unused.php` — we end up with a custom configuration file in the final application. I think it would be better to generate it on the fly. If you’re willing to reopen this against Sylius 2.1 (or 2.2 if it’s already been reopened), please go ahead :)

PR #694, opened 2022-04-06
[README] Add reasonable defaults configs for mailing and messenger to avoid production errors

lchrusciel:
The current mailing config results in 500 errors - this is the result that we can be sure of. So, it is better to turn it off by default, in my humble opinion. People who need mailing will adjust it anyway. The rest won't have a broken website. 

The second config is quite similar. This is our recommendation, so it is reasonable to have it there.


AdamKasp (approved):

GSadee (approved):

Roshyo (approved):

Roshyo:
But doesn't the messenger part require a command line to install and configure DB tables ?

lchrusciel:
@Roshyo we are fine. It may be done by command line, but Symfony messenger will take care of it during initialization

Zales0123 (approved):

Zales0123:
Thank you, Łukasz! :tada:

Issue #692, opened 2022-04-05
Sylius dependency problem in branches 1.10 and 1.11


delyriand:
The version 1.10 of sylius-standard installs the last version of sylius 1 (currently 1.11) instead of the version 1.10.

This is because you have a "caret version range" on the sylius requirement in [composer.json](https://github.com/Sylius/Sylius-Standard/blob/1.10/composer.json#L29):
```json
    "require": {
        // ...
        "sylius/sylius": "^1.10",
        // ...
    },
```
Is equivalent of `>=1.10 <2.0.0`. 

The 1.9 version does't have the issue, because you use the "tilde version range":
```json
    "require": {
        // ...
        "sylius/sylius": "~1.9.0",
        // ...
    },
```
Is equivalent of `>=1.9.0 <1.10.0`.

**Possible solution:**

Correct the dependency constraint with a title:
- for branch 1.10: `~1.10.0`
- for branch 1.11: `~1.11.0`

delyriand:
And we have an error when we create a project with the sylius-standard 1.10:

```
!!  In CheckExceptionOnInvalidReferenceBehaviorPass.php line 86:
!!
!!    The service "sylius.shipping_date_assigner" has a dependency on a non-exist
!!    ent service "Sylius\Calendar\Provider\DateTimeProviderInterface".
```

I think it's a configuration error between sylius 1.10 - the expected version - and sylius 1.11 - the installed version.

tbreton14:
Yes i confirm, i have the same problem !

PR #701, opened 2022-04-12
Fix 1.10 version problem

Zales0123:
Indeed, the constraint of the `sylius/sylius` package in the composer.json allows for too much 🖖 I believe we can live it as it is in 1.11 and when the 1.12 version would be released we will change it to ~1.11.0 as well. 
I've also improved Github actions configuration a little bit to test more than one of the supported Symfony versions  🚀 

Fixes https://github.com/Sylius/Sylius-Standard/issues/692


GSadee (approved):

GSadee:
Thanks, Mateusz! :tada:

PR #719, opened 2022-05-04
[docker]Remove Docker Compose optional version

Ferror:
Screenshot 2022-05-04 at 09 22 12

The Docker Compose `version` is optional and there is no value to keeping that as the latest specification is just a "Compose Specification" instead of value-versioning https://docs.docker.com/compose/compose-file/

More to read in Docker Compose versioning reference: https://docs.docker.com/compose/compose-file/compose-versioning/

Current Docker Compose version: v2.4.1

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,5 +1,3 @@
-version: '3.4'
-
 services:
   php:
     image: quay.io/sylius/php:latest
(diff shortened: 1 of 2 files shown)


lchrusciel (approved):

lchrusciel:

PR #720, opened 2022-05-04
[docker]Execute yarn build in production Dockerfile

Ferror:
The Dockerfile is "production oriented", yet it specifies the default command as `yarn watch` which executes a process of actively looking for changes. We don't need to do that in the production environment as the purpose of the node container would be to build and store assets and media files.

To maintain development environment the `yarn watch` command was specified in the `docker-compose`


lchrusciel (approved):

lchrusciel:
Thanks, Zbigniew! :tada:

PR #725, opened 2022-05-05
[docker]Use mysql instead of percona

Ferror:
The Percona was introduced in the docker-compose in 2018, but we officially do not support this database provider (even if it is MySQL-based db engine)

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -20,7 +20,7 @@ services:
 
   mysql:
     # in production, we may want to use a managed database service
-    image: percona:5.7
+    image: mysql:5.7 # Sylius is fully working on mysql 8.0 version
     environment:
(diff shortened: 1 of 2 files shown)


docker-compose.prod.yml
```diff
@@ -20,7 +20,7 @@ services:
 
   mysql:
     # in production, we may want to use a managed database service
-    image: percona:5.7
+    image: mysql:5.7 # Sylius is fully working on mysql 8.0 version
```
lchrusciel, line 23: Why didn't you change the version to 8.0?

Ferror: There is no reason. We support both versions, but which is preferable for us - I don't know.

I can create additional PR with the version change, as the goal of PR was to clarify the db engine. Comment was extra 😄

lchrusciel (approved):

lchrusciel:

PR #690, opened 2022-04-01
Remove usage of libsass

Nek-:
Follow up https://github.com/Sylius/Sylius/pull/13488

PR #728, opened 2022-05-09
[Maintenance] Update Sylius to newest dev master

lchrusciel:
From time to time, the last released version is not enough on master here. Some changes we would like to prepare in advance like #690. 

^1.12@dev makes the most sense to me - it is clear, that we are working on the development of v1.12, but accepting dev-master. 

Ref. https://semver.madewithlove.com/?package=sylius%2Fsylius&constraint=^1.11&stability=dev (but for version 1.11 to see how it behaves)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,7 +129,7 @@ jobs:
                 name: Prepare test application assets
                 run: |
                     bin/console assets:install public -vvv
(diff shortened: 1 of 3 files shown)


Zales0123 (approved):

Zales0123:
Thank you @lchrusciel and @Nek-! 🏅

PR #759, opened 2022-05-27
Pipeline that will test Sylius Store on Docker

Ferror:
The pipeline will build Sylius Store from `docker-compose.yml` and `docker-compose.prod.yml`, but only on the linux/amd64 platform as there is no functionality to override platform from the command nor does GitHub provide ARM runners.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,9 +6,11 @@
 **/.DS_Store
 **/.git/
 **/.gitattributes
+**/.github
 **/.gitignore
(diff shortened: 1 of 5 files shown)


lchrusciel (approved):
Why did you call it Sylius-Store?

Ferror:
> Why did you call it Sylius-Store?

I thought that it is the right name for the... instance of Sylius-Standard. Do you find it misleading or would you recommend a different name?

PR #800, opened 2022-07-19
[docker] Create Sylius Plus authentication placeholder via COMPOSER_AUTH env

Ferror:


Ferror:
We cant do it, as the ENV is not used during the build.

PR #812, opened 2022-08-15
[docker]Create container that runs cronjobs

Ferror:
With the docs: https://docs.sylius.com/en/latest/cookbook/deployment/cron-jobs.html comes requirements or a feature to run cronjobs for Sylius. The current docker does not provide this functionality out of the box, yet it would be nice. There it is.

The cronjobs are not helpful in the dev environments because containers tend to die rather quickly and commands are triggered manually. That's why the cronjobs container applies only to the production version of docker-compose.

### DIFF
--- a/docker/cron/crontab
+++ b/docker/cron/crontab
@@ -1,0 +1,2 @@
+* * * * * bash php bin/console sylius:remove-expired-carts
+* * * * * bash php bin/console sylius:sylius:cancel-unpaid-orders
--- a/docker/cron/crontab
+++ b/docker/cron/crontab
@@ -1,2 +1,2 @@
-* * * * * bash php bin/console sylius:remove-expired-carts
-* * * * * bash php bin/console sylius:sylius:cancel-unpaid-orders
+* * * * * sh php bin/console sylius:remove-expired-carts
+* * * * * sh php bin/console sylius:sylius:cancel-unpaid-orders
--- a/Dockerfile
+++ b/Dockerfile
@@ -128,3 +128,15 @@ ARG APP_ENV=dev
 RUN set -eux; \
     composer install --prefer-dist --no-autoloader --no-interaction --no-scripts --no-progress; \
     composer clear-cache
+
+FROM sylius_php_prod AS sylius_cron
+
+RUN set -eux; \
+	apk add --no-cache --virtual .build-deps \
+		apk-cron \
+	;
+
+COPY docker/cron/crontab /etc/crontabs/root
+
+ENTRYPOINT ["crond"]
+CMD ["-f"]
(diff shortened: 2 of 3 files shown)


docker/cron/crontab
```diff
@@ -0,0 +1,2 @@
+* * * * * sh php bin/console sylius:remove-expired-carts
+* * * * * sh php bin/console sylius:sylius:cancel-unpaid-orders
```
coldic3, line 2:
```suggestion
0 */6 * * * sh php bin/console sylius:remove-expired-carts
0 */6 * * * sh php bin/console sylius:sylius:cancel-unpaid-orders
```

Probably it doesn't have to be that frequent. The value I propose is copied from the docs: https://docs.sylius.com/en/1.12/cookbook/deployment/platform-sh.html#add-default-sylius-cronjobs

Ferror: I know why I did'nt see that. I focused on this page: https://docs.sylius.com/en/1.12/cookbook/deployment/cron-jobs.html?highlight=cron#how-to-configure-a-cron-job

lchrusciel: we should also add Symfony messenger for catalog promotions
```suggestion
* * * * * sh php bin/console sylius:remove-expired-carts
* * * * * sh php bin/console sylius:sylius:cancel-unpaid-orders
* * * * * sh php bin/console messenger:consume main
```

coldic3: 👍️ and we can add `--time-limit=60` flag to the `messenger:consume` command to have only one consumer running in the background. However, well configured supervisor would be better in this case, but for now, cron job is better than nothing 😛

Ferror: Valid point about the catalog promotions, but yea The messenger:consume will have separated queue worker. That's my next step :)

docker/cron/crontab
```diff
@@ -0,0 +1,2 @@
+0 */6 * * * sh php bin/console sylius:remove-expired-carts
```
lchrusciel: Shouldn't we add catalog promotion processing here as well?

Ferror: I thought I responded in the previous thread. As far as I know the catalog promotions are executed via symfony messenger. Therefore it should be run a little bit different - by a worker (bin/console messenger:consume) and it should be separated container. If I'm wrong - let me know

coldic3 (approved):

lchrusciel (approved):

docker/cron/crontab
lchrusciel: you've answered but also resolved comment so I didn't see it. Seems legit 👍

PR #816, opened 2022-08-24
[GH] Run builds on latest ubuntu

lchrusciel:


Ferror:
We change it to ubuntu-20.04 LTS version

Issue #813, opened 2022-08-19
docker-compose.prod.yml reload fixture on docker restart


duchnoun:
Hi,

Can't understand why fixture are reload (and delete all data) if container restart. (In case of docker-compose.prod.yml)

In documentary we launch  : docker compose -f docker-compose.prod.yml exec php bin/console sylius:fixtures:load --no-interaction
Ok it's normal the first time if we want fixture.

But why it load it on : https://github.com/Sylius/Sylius-Standard/blob/1.12/docker/php/docker-entrypoint.sh (line 26)

In production data shouldn't be erase ? Or i mistake ? 

Thanks

duchnoun:
no body got a response ? thanks :)

vvasiloi:
@Ferror looks like your work.

Ferror:
@duchnoun The docker-entrypoint.sh is shared for all environments. Why did I add the fixture loading? It's much better DX at the beginning. When you start a project and see a proper store instead of an error page with information that the "Channel was not found". If it does not suit your needs - feel free to remove it for your project 😃 

PS. Thx @vvasiloi for a ping 😄

vvasiloi:
@Ferror the intention was good, but it ended up being a bad DX in a different way.
I suggest you do that conditionally.
Don't do it if env is prod and add a state for initialization.
For the state, a common approach is to mount a dir or create a file after the initilization.
Another option is to run `bin/console doctrine:query:sql` and check if channels exists.
Last but not least, make it possible to easily opt-out, with an env var for example.

duchnoun:
@vvasiloi i agree with you +1.

vasilvestre:
IMO and as dunglas does in his docker, there's a comment that says "after first install feel free to remove this"
The first install is simpler with this, people should edit files to match the behaviour they want to have

Ferror:
In https://github.com/Sylius/Sylius-Standard/pull/859/files we introduced an environment variable that can easily turn loading fixtures off

PR #826, opened 2022-09-15
[docker] Load fixtures only on dev environment

Ferror:
https://github.com/Sylius/Sylius-Standard/issues/813

Turning off relading the fixtures is just one line, and I decided not to overcomplicate the default config and just keep the fixtures reloading on the dev environment, but not on production and testing.

### DIFF
--- a/docker/php/docker-entrypoint.sh
+++ b/docker/php/docker-entrypoint.sh
@@ -23,7 +23,10 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'bin/console' ]; then
 	done
 
     bin/console doctrine:migrations:migrate --no-interaction
-    bin/console sylius:fixtures:load --no-interaction
+
+    if [ "$APP_ENV" = 'dev' ]; then
+        bin/console sylius:fixtures:load --no-interaction
+    fi
 fi
 
 exec docker-php-entrypoint "$@"


vvasiloi:
Now run `docker-compose run --rm php bin/console any:command` and see what happens.

Ferror:
> Now run `docker-compose run --rm php bin/console any:command` and see what happens.

Why would you do `docker compose run` instead of `docker compose exec`? The container is up and running. Btw. Even without this change the service would start and execute everything that is in the `docker-entrypoint.sh` file. So I don't understand the point 😢 

If you would like to use docker compose run... we should get rid of one of the if statements, but I don't believe that's the way the author wanted to use the container.

vvasiloi:
The point is that the dev should be able to use `docker-compose run php`, `docker-compose restart php`, `docker-compose up -d --scale php=3` 10 times a day without resetting the database every time.
Let's say I set up a context for reproducing a bug or testing a feature, I might add tons of data, and then for some reason I have to restart the container. All my work is gone now.
I always made my own Docker setups for Sylius, so I don't care about it for myself., but as more people will start using it I'm sure more will complain, it's just a matter of time.

Ferror:
> The point is that the dev should be able to use `docker-compose run php`, `docker-compose restart php`, `docker-compose up -d --scale php=3` 10 times a day without resetting the database every time. Let's say I set up a context for reproducing a bug or testing a feature, I might add tons of data, and then for some reason I have to restart the container. All my work is gone now. I always made my own Docker setups for Sylius, so I don't care about it for myself., but as more people will start using it I'm sure more will complain, it's just a matter of time.

I think you may want to convert that message to the issue, as there is a specific thing you want to achieve and it's more like feedback for the whole PHP container execution than this particular change.

Ferror:
It will be changed by other PR. https://github.com/Sylius/Sylius-Standard/pull/859

Issue #832, opened 2022-09-30
Fresh install in env PROD with docker fail


duchnoun:
Hi,

On a fresh system (clear all image , old build ect..)

----
git clone https://github.com/Sylius/Sylius-Standard.git
cd Sylius-Standard
export MYSQL_PASSWORD=SLyPJLaye7
docker-compose -f docker-compose.prod.yml up -d
docker-compose exec php bin/console sylius:fixtures:load --no-interaction
------

After visiting the website : error 500.

log 👍 

[2022-09-30T10:52:52.890581+00:00] request.CRITICAL: Exception thrown when handling an exception (Twig\Error\RuntimeError: An exception has been thrown during the rendering of a template ("Could not find the entrypoints file from Webpack: the file "/srv/sylius/public/build/shop/entrypoints.json" does not exist."). at /srv/sylius/templates/bundles/SyliusShopBundle/_styles.html.twig line 1) {"exception":"[object] (Twig\\Error\\RuntimeError(code: 0): An exception has been thrown during the rendering of a template (\"Could not find the entrypoints file from Webpack: the file \"/srv/sylius/public/build/shop/entrypoints.json\" does not exist.\"). at /srv/sylius/templates/bundles/SyliusShopBundle/_styles.html.twig:1)\n[previous exception] [object] (InvalidArgumentException(code: 0): Could not find the entrypoints file from Webpack: the file \"/srv/sylius/public/build/shop/entrypoints.json\" does not exist. at /srv/sylius/vendor/symfony/webpack-encore-bundle/src/Asset/EntrypointLookup.php:124)"} []

--- in dev mode it work .. ---

if i check in docker php container there is no /srv/sylius/public/build folder. (but it exist on the host)

Where i mistake ?

Thanks.

duchnoun:
no body got a response ? thanks :)

vvasiloi:
Run `yarn install` and `yarn build` before that.

vvasiloi:
@Ferror got an idea how to solve this?

Ferror:
I just got aware of the issue. It's because of the PR: https://github.com/Sylius/Sylius-Standard/pull/831 I will try to pull off the changes

vvasiloi:
@Ferror I don't understand how is that PR related to this issue.

Ferror:
Well simply as it gets the entires with webpack inside the docker does not load/generate properly

vvasiloi:
[...]

PR #831, opened 2022-09-28
[Maintenance] Prepare Sylius-Standard for Sylius 1.12

jakubtobiasz:
This PR should also close #828.

PR #833, opened 2022-10-03
[docker]Keep the Gulp in the docker and prepare proper migration

Ferror:
As we are going to support both GULP and WebPack in Sylius 1.12 we should do the same in Docker containers. I reverted some of the changes from https://github.com/Sylius/Sylius-Standard/pull/831 PR because the main branch docker is broken.

Let's make a step back and prepare the node container for webpack and for gulp.

FYI: https://github.com/Sylius/Sylius-Standard/issues/832


Ferror:
F. it. The webpack is required.

PR #835, opened 2022-10-03
[docker]Integrate production config docker with Webpack

Ferror:
FYI https://github.com/Sylius/Sylius-Standard/issues/832

The PHP container now needs the frontend files. This is not the most distinguished solution, but it works. I will spend some more time polishing it. But now it is required to fix the 1.12 branch.

PS. it worked on the development image only because the command generated the files and syncs with local filesystem

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -25,6 +25,7 @@ node_modules/
 var/*
 vendor/
 public/assets/
+public/build/
 public/bundles/
 public/css/
 public/js/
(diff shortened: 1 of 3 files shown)


vvasiloi (approved):

jakubtobiasz (approved):

jakubtobiasz:

duchnoun:
Thanks :)

PR #842, opened 2022-10-13
[docker] Simple queue worker

Ferror:
The ideal scenario would be to have PHP CLI image that would run as a worker consumer, but It would increase the complexity. Let's start with a simple worker that would consume messages and always be up with the `restart: always` docker composer directive.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -7,12 +7,13 @@ services:
     depends_on:
       - mysql
     environment:
+      APP_DEBUG: 0
+      APP_ENV: prod
+      APP_SECRET: EDITME
       DATABASE_URL: mysql://sylius:${MYSQL_PASSWORD}@mysql/sylius_prod
       MAILER_URL: smtp://localhost
+      MESSENGER_TRANSPORT_DSN: doctrine://default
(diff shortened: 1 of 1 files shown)


coldic3:
IMO there are few problems with that solution. Firstly we have only one worker running while it would be nice to have multiple of them on production. We also haven't `--limit` and `--memory-limit` options specified so the worker might consume massive amounts of memory and won't be restarted. WDYT?

Ferror:
> IMO there are few problems with that solution. Firstly we have only one worker running while it would be nice to have multiple of them on production. We also haven't `--limit` and `--memory-limit` options specified so the worker might consume massive amounts of memory and won't be restarted. WDYT?

Alright. So I added the limit and memory etc for workers.

memory-limit - default to php.ini 256MB
limit - I found it comment to use 10 or 5 per worker.
time-limit - Each message should easily be consumed under 1 minute, so I specified max as 10 minutes per whole 5 messages

Regarding scaling. Well, Docker Compose is not the orchestrator that's why we won't be able to scale anything here unless we turn the swam mode. https://docs.docker.com/engine/reference/commandline/service_scale/

coldic3 (approved):

lchrusciel (approved):

PR #844, opened 2022-10-13
[docker] Update container's dependencies versions

Ferror:
Specify a minor version for the composer container. Update PHP to 8.1. Therefore we can get rid of php81 polyfill.

Also cleaned up files and extensions. Some of them are already installed. And we don't need .env.test in production container


jakubtobiasz (approved):

GSadee (approved):

Rafikooo (approved):

ernestWarwas:
Thank you, Zbigniew! :tada:

PR #858, opened 2022-10-26
[Composer] Sylius and SyliusResourceBundle versions adjusted

Rafikooo:
The order of installation of dependencies results in an outdated `SyliusResourceBundle` package which prevents adding products to the cart.
Requiring the latest `SyliusResourceBundle` adjusts the `phpstan/doc-parser` version:
image
The requirement of an explicit version of `SyliusResourceBundle` should be removed after the release of `Sylius 1.12`

It also fixes an installation because of that change: https://github.com/Sylius/PayPalPlugin/pull/266

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     "require": {
         "php": "^8.0",
         "sylius/paypal-plugin": "^1.5.x-dev",
-        "sylius/sylius": "^1.12.x-dev",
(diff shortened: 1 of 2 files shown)


Rafikooo:
https://github.com/Sylius/PayPalPlugin/pull/267 resolves this issue

PR #859, opened 2022-10-31
[docker] - Create migration container

Ferror:
Flow Before

image

Flow After
image


docker-compose.prod.yml
```diff
@@ -34,6 +50,7 @@ services:
       APP_ENV: prod
       APP_DEBUG: 0
       APP_SECRET: EDITME
+      LOAD_FIXTURES: 1
```
coldic3:
```suggestion
      LOAD_FIXTURES: 0
```
Is there a reason why it is enabled on production?

docker-compose.yml
```diff
@@ -22,6 +22,21 @@ services:
...
 
+  migrations:
+    container_name: migrations
+    build:
+      context: .
+      target: sylius_migrations
+    depends_on:
+      - mysql
+    environment:
+      - APP_ENV=prod
```
coldic3:
```suggestion
      - APP_ENV=dev
```

docker-compose.prod.yml
Ferror: I still want to preserve the flow where Sylius shows an excellent shop front by default. But I got feedback that it may be weird from DX perspective. That's why I kept this easy switch.

Without loaded fixtures user will get error page by default - that's blah :(

Ferror:
PS. after this PR I will start working on the worker, cron, and migrations container to not have PHP-FPM. Those are going to be CLI-based containers

docker-compose.prod.yml
coldic3: Hmm, I see but on the other hand, it removes all data from the database in every single build which is also not ok. As it is production, IMO we shouldn't care about the error page because the admin can configure everything including channels.
Nevertheless, it's a docker compose file that can be overwritten so I'm ok with it as it is 👍️

coldic3 (approved):

docker-compose.prod.yml
Ferror: I will think about it. Like I get pros and cons and maybe I should give it a try ;p

Ferror: I decided to give it a try :)

jakubtobiasz (approved):

TheMilek (approved):

GSadee (approved):

NoResponseMate (approved):

ernestWarwas (approved):

GSadee:
Thank you, Zbigniew! :tada:

PR #863, opened 2022-11-01
[maintenance] - Store sessions in database instead of files

Ferror:
Based on Symfony Docs:

https://symfony.com/doc/current/session/database.html#store-sessions-in-a-relational-database-mariadb-mysql-postgresql

Storing sessions in files, especially locally prevents us from scaling Sylius because it is additional calls to the file system that is shared between multiple machines. Ideally we would like to store session in Redis or other Key-Value database but to no add additional service to our stack we keep it simple - in database.


coldic3 (approved):

jakubtobiasz (approved):

src/Migrations/Version20221101193636.php
```diff
@@ -0,0 +1,33 @@
...
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20221101193636 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $sql = <<<SQL
+CREATE TABLE sessions (
```
lchrusciel, line 15: don't we need some config like this? 
```yaml
doctrine:
    dbal:
        connections:
            default:
                schema_filter: ~^(?! sessions)~
```

PR #871, opened 2022-11-10
Use @sylius-ui/frontend npm package

jakubtobiasz:
Following https://github.com/Sylius/Sylius/pull/14529 and https://github.com/Sylius/Sylius/issues/12132.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /public/assets
+/public/build
 /public/css
 /public/js
 /public/media
--- a/package.json
+++ b/package.json
@@ -1,68 +1,13 @@
 {
-  "dependencies": {
-    "@babel/polyfill": "^7.0.0",
-    "chart.js": "^3.7.1",
-    "jquery": "^3.5.0",
-    "jquery.dirtyforms": "^2.0.0",
-    "lightbox2": "^2.9.0",
-    "semantic-ui-css": "^2.2.0",
-    "slick-carousel": "^1.8.1"
-  },
-  "devDependencies": {
-    "@babel/core": "^7.0.0",
-    "@babel/plugin-external-helpers": "^7.0.0",
-    "@babel/plugin-proposal-object-rest-spread": "^7.18.9",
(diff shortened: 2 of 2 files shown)


vvasiloi:
Since all `dependencies` are now `devDependencies`, are there any limitations users should be aware of?

jakubtobiasz:
Nope, the whole application is built, so none dependency is used later.

Ferror:
I don't understand why the Sylius frontend is dev dependency

jakubtobiasz:
@Ferror notice all gulp, webpack, and others were in dev section. None of these packages are required by runtime, as they all are only used to build the result files. All of these deps can be placed in both dependencies and devDependencies (I mean the result will be the same) so as more deps were in dev then I placed this package as dev dependency too.

Ferror:
> @Ferror notice all gulp, webpack, and others were in dev section. None of these packages are required by runtime, as they all are only used to build the result files. All of these deps can be placed in both dependencies and devDependencies (I mean the result will be the same) so as more deps were in dev then I placed this package as dev dependency too.

Yea they were all in devDependency and then run with GULP_ENV=prod.

jakubtobiasz:
> > @Ferror notice all gulp, webpack, and others were in dev section. None of these packages are required by runtime, as they all are only used to build the result files. All of these deps can be placed in both dependencies and devDependencies (I mean the result will be the same) so as more deps were in dev then I placed this package as dev dependency too.
> 
> Yea they were all in devDependency and then run with GULP_ENV=prod.

Now they are `yarn build:prod` what under the hood is `yarn encore production`, I don't see any difference.

coldic3 (approved):

NoResponseMate (approved):

Rafikooo (approved):
I'm wondering if we should open the new branch (`1.13-dev`) as [these](https://github.com/Sylius/Sylius/pull/14529) changes are targeting `Sylius 1.13`

Rafikooo:
Thank you, Jakub! :1st_place_medal:

Issue #872, opened 2022-11-15
How about removing `yarn.lock` from repository?


Ferror:
We don't store `composer.lock` file that locks PHP dependencies. Maybe we should do the same with Javascript dependencies?

PR #855, opened 2022-10-21
[Composer] Remove composer.lock from .gitignore after creating the project

artyuum:
This fixes the issue #665.

Inspired by [sulu/skeleton](https://github.com/sulu/skeleton/blob/9f6f4d5792a737e96e20b5d9cce9523e43e064b7/composer.json#L118).

Can you review this, please? @lchrusciel

PR #873, opened 2022-11-15
[maintenance] - Ignore yarn.lock file

Ferror:
PR Related with https://github.com/Sylius/Sylius-Standard/issues/872


vvasiloi:
We should probably think about automatically removing it from .gitignore when a new project is created.
Here's a proposed solution for the composer.lock: https://github.com/Sylius/Sylius-Standard/pull/855

vvasiloi (approved):

Dockerfile
```diff
@@ -93,7 +93,7 @@ RUN set -eux; \
 	;
 
 # prevent the reinstallation of vendors at every changes in the source code
-COPY package.json yarn.lock ./
+COPY package.json yarn.* ./
```
coldic3, line 96:
```suggestion
COPY package.json yar[n].lock ./
```
is a safer option to avoid having errors for a non-existent file but it looks strange 😄

coldic3 (approved):

PR #877, opened 2022-11-25
[Maintenance] `sylius/sylius-rector` instead of `rector/rector`

Rafikooo:


Zales0123 (approved):

GSadee (approved):

composer.json
```diff
@@ -52,9 +52,9 @@
         "phpstan/phpstan-webmozart-assert": "^1.1",
         "phpunit/phpunit": "^8.5",
         "polishsymfonycommunity/symfony-mocker-container": "^1.0",
-        "rector/rector": "^0.14.6",
         "stripe/stripe-php": "^6.43",
         "sylius-labs/coding-standard": "^4.0",
+        "sylius/sylius-rector": "0.11.x-dev",
```
GSadee, line 57: Why do we have `0.11.x-dev` version there?

GSadee:
Thank you, Rafał! :1st_place_medal:

composer.json
Rafikooo: I think both `0.11.x-dev` and `dev-main` are correct as `0.11-dev` is set as an alias

GSadee: My point was that SyliusRector has no released version yet, so why is it targeting 0.11 instead of 0.1 (or possibly 1.0)?

Rafikooo: I relied on [this](https://github.com/Sylius/SyliusRector/blob/main/composer.json#L48). If it's wrong I'll change it

PR #887, opened 2022-12-08
[Docker] Fix fixture images not being available #885

### LINKED ISSUE #885: Cannot resolve images shipped with fixtures
## Environement

Fresh installation of Sylius-Standard by docker on 7.12.2022. Development and production environment.

## Introduction

With adding webpack to the Sylius and Sylius-Standard the problem of missing thumbnails generated by liip package occurred. I don't know what the reason behind it is, but I would like to document the issue. Based on Sylius Community Slack the problem does exist without the usage of docker too.

Screenshot 2022-12-07 at 23 01 17

Screenshot 2022-12-07 at 23 00 11

## Example

We have `http://localhost/media/cache/resolve/sylius_shop_product_thumbnail/f4/cf/35bf6a62b06e31be473ac49d0ec9.jpg` image` it does NOT exist in `public/media` directory for Nginx and PHP containers. Well, the directory is empty.

Therefore we are missing some steps in Docker or Webpack.

The file responsible for loading the image -> https://github.com/Sylius/Sylius/blob/1.12/src/Sylius/Bundle/ShopBundle/Resources/views/Product/_mainImage.html.twig

## Logs

```
[2022-12-07T22:33:21.964352+00:00] request.ERROR: Uncaught PHP Exception Symfony\Component\HttpKernel\Exception\NotFoundHttpException:
"Source image for path "6b/fd/1c6ee947b9a6cd20d5599be9322a.jpg" could not be found"
at /srv/sylius/vendor/liip/imagine-bundle/Controller/ImagineController.php line 168 {"exception":"[object] (Symfony\\Component\\HttpKernel\\Exception\\NotFoundHttpException(code: 0): Source image for path \"6b/fd/1c6ee947b9a6cd20d5599be9322a.jpg\" could not be found at /srv/sylius/vendor/liip/imagine-bundle/Controller/ImagineController.php:168)

[previous exception] [object] (Liip\\ImagineBundle\\Exception\\Binary\\Loader\\NotLoadableException(code: 0): Source image not resolvable \"6b/fd/1c6ee947b9a6cd20d5599be9322a.jpg\" in root path(s) \"/srv/sylius/public/media/image\" at /srv/sylius/vendor/liip/imagine-bundle/Binary/Locator/FileSystemLocator.php:47)"} []

```

vvasiloi:
Fix #885

1. mount media volume to migrations container so fixture images are available in php and nginx container
2. change nginx user and group to match php's, so it can read image cache files

### DIFF
--- a/docker/migrations/docker-entrypoint.sh
+++ b/docker/migrations/docker-entrypoint.sh
@@ -22,4 +22,7 @@ php bin/console doctrine:migrations:migrate --no-interaction
 
 if [ "$LOAD_FIXTURES" = "1" ]; then
     php bin/console sylius:fixtures:load --no-interaction
+
+    # make the image files created by fixtures accessible by fpm which runs with user www-data
+    chown -R www-data:www-data public/media/image
 fi
--- a/docker/nginx/nginx.conf
+++ b/docker/nginx/nginx.conf
@@ -1,0 +1,31 @@
+user  www-data;
+worker_processes  auto;
+
+error_log  /var/log/nginx/error.log notice;
+pid        /var/run/nginx.pid;
+
+
+events {
+    worker_connections  1024;
+}
+
+
+http {
+    include       /etc/nginx/mime.types;
+    default_type  application/octet-stream;
+
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                      '$status $body_bytes_sent "$http_referer" '
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    keepalive_timeout  65;
+
+    #gzip  on;
+
+    include /etc/nginx/conf.d/*.conf;
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -118,6 +118,15 @@ COPY --from=sylius_node /srv/sylius/public/build public/build
 
 FROM nginx:${NGINX_VERSION}-alpine AS sylius_nginx
 
+ARG GID=82
+ARG UID=82
+ARG USERNAME="www-data"
+
+RUN set -x; \
+    addgroup -g $GID -S $USERNAME; \
+    adduser -S -D -H -u $UID -h /var/cache/nginx -s /sbin/nologin -G $USERNAME -g $USERNAME $USERNAME
+
+COPY docker/nginx/nginx.conf /etc/nginx/nginx.conf
 COPY docker/nginx/conf.d/default.conf /etc/nginx/conf.d/
 
 WORKDIR /srv/sylius
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,6 +36,14 @@ services:
       - DATABASE_URL=mysql://sylius:${MYSQL_PASSWORD:-nopassword}@mysql/sylius
       - LOAD_FIXTURES=1
       - PHP_DATE_TIMEZONE=${PHP_DATE_TIMEZONE:-UTC}
+    volumes:
+      - .:/srv/sylius:rw,cached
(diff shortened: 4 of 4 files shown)


docker/nginx/nginx.conf
```diff
@@ -0,0 +1,31 @@
+user  www-data;
```
Ferror, line 1: Do you think we can merge both configurations?

https://github.com/Sylius/Sylius-Standard/blob/1.12/docker/nginx/conf.d/default.conf

if they can be separate, then I would rather have them moved to conf.d directory and let nginx handle loading it

Ferror:
I can see that the image files are now available, but the requests are returning 404. Fresh Sylius-Standard installation.

Screenshot 2022-12-08 at 23 09 50

Screenshot 2022-12-07 at 23 00 11

docker/nginx/nginx.conf
vvasiloi: The files from `conf.d` are included in the `http` section, they can't be merged and this file can't go to `conf.d`.

docker/nginx/nginx.conf
```diff
@@ -0,0 +1,31 @@
...
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    keepalive_timeout  65;
+
+    #gzip  on;
+
+    include /etc/nginx/conf.d/*.conf;
```
vvasiloi, line 30: @Ferror this is where the files from `conf.d` are included. This config file is a copy of the one that was shipped with this image, I just changed the user.

docker/nginx/nginx.conf
vvasiloi: It's how docs recommend to customize it: https://github.com/docker-library/docs/tree/master/nginx#complex-configuration

vvasiloi:
I hate root containers... the problem still persist becuase the migrations container runs as root and the image files are created with root user, but now the nginx and fpm process run with a non-root user...

docker/migrations/docker-entrypoint.sh
```diff
@@ -22,4 +22,7 @@ php bin/console doctrine:migrations:migrate --no-interaction
 
 if [ "$LOAD_FIXTURES" = "1" ]; then
     php bin/console sylius:fixtures:load --no-interaction
+
+    # make the image files created by fixtures accessible by fpm which runs with user www-data
+    chown -R www-data:www-data public/media/image
```
vvasiloi, line 27: @Ferror this will fix the issue, but it's just like applying FlexTape... ideally we should adopt non-root containers and make GIDs and UIDs configurable via build args and env vars, like this: https://github.com/laradock/laradock/blob/96e0f2e92f6e5b128a71cc77a05b297ac4a0eadb/workspace/Dockerfile#L43-L47

Ferror:
PS. The pipeline probably fails because MySQL is taking a little more time to spin up. Let's try to increment the attempt amount to 25. It's still gonna be only 25 seconds in the worst-case scenario, and in real life, the database would be always available.

https://github.com/Sylius/Sylius-Standard/blob/1.12/docker/migrations/docker-entrypoint.sh#L4

Ferror (approved):

umpirsky:
I have the same issue. Can we merge this?

docker/migrations/docker-entrypoint.sh
briais-s: I realized that if I launched the fixtures manually from the php container, I had to reset the rights each time afterwards.

```
php bin/console sylius:fixtures:load -n && chown -R www-data:www-data public/media/image
```

gabiudrescu:
@vvasiloi who do we need to bribe to get this merged in master?

vvasiloi:
> @vvasiloi who do we need to bribe to get this merged in master?

This should not be merged. It's not a good solution. We should use non-root containers instead.

PR #889, opened 2022-12-09
[Docker] Change wildcard to safely copy yarn.lock only

coldic3:
see https://github.com/Sylius/Sylius-Standard/pull/873#discussion_r1027736382

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,7 +93,7 @@ RUN set -eux; \
 	;
 
 # prevent the reinstallation of vendors at every changes in the source code
-COPY package.json yarn.* ./
+COPY package.json yar[n].lock ./
 RUN set -eux; \
     yarn install; \
     yarn cache clean


Dockerfile
```diff
@@ -93,7 +93,7 @@ RUN set -eux; \
 	;
 
 # prevent the reinstallation of vendors at every changes in the source code
-COPY package.json yarn.* ./
+COPY package.json yar[n].lock ./
```
Rafikooo, line 96: Could you explain why `[n]`?

Rafikooo (approved):

Dockerfile
coldic3: This is a wildcard that accepts "n" char and only "n" char XD I know it's a bit hacky but this workaround allows us to copy `yarn.lock` if exists or skip it if not. Without any wildcard (previously wildcard was an asterisk), building the docker image fails because can't find the `yarn.lock` file.

Rafikooo: Indeed, it's tricky, but now I understand, thanks :D

rust-le:

PR #865, opened 2022-11-03
[Maintenance] Annotation changed to PHP Attributes in the Entity classes

Rafikooo:
Since we do not use PHP 7.4 anymore, it's preferable to move into PHP Attributes

PR #891, opened 2022-12-13
[Maintenance] PHP Attributes in the Entity classes

Rafikooo:
The follow-up to: https://github.com/Sylius/Sylius-Standard/pull/865 & https://github.com/Sylius/Sylius-Standard/pull/881

Since we do not use PHP 7.4 anymore, it's preferable <s>to move into PHP Attributes</s> to support both annotations and attributes with default annotations setting 🎉

### DIFF
--- a/src/Entity/Addressing/Address.php
+++ b/src/Entity/Addressing/Address.php
@@ -11,6 +11,8 @@ use Sylius\Component\Core\Model\Address as BaseAddress;
  * @ORM\Entity
  * @ORM\Table(name="sylius_address")
  */
+#[ORM\Entity]
+#[ORM\Table(name: 'sylius_address')]
 class Address extends BaseAddress
 {
 }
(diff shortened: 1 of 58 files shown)


NoResponseMate (approved):
You can also add it as the default type in [doctrine's config](https://github.com/Sylius/Sylius-Standard/blob/1.12/config/packages/doctrine.yaml).
If someone prefers to still use annotations, they can easily change it.

Rafikooo:
@NoResponseMate, as I recall, we agreed that in Sylius 1.12 we will stick with annotations, giving the ability to switch to the attributes, but we will use attributes as the default configuration with Sylius 1.13

jakubtobiasz (approved):
I don't understand why we provide it that way, when there's a Rector rule https://getrector.org/blog/how-to-upgrade-annotations-to-attributes (Rector is installed by default since 1.12 AFAIR).

If we really need it to have declared, I'm fine with that 👌🏼.

jakubtobiasz:
Thanks, Rafał! :tada:

PR #896, opened 2023-02-01
[Maintenance] Add missing messenger env vars

coldic3:
This change is due to https://github.com/Sylius/Sylius/pull/14716


.env
```diff
@@ -17,8 +17,12 @@ DATABASE_URL=mysql://root@127.0.0.1/sylius_%kernel.environment%
 ###> symfony/messenger ###
 # Choose one of the transports below
 # MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages
-MESSENGER_TRANSPORT_DSN=doctrine://default
```
Rafikooo, line 20: We only removed it here. I think it's safe to remove it across the rest of the configuration files to let the new projects know that `MESSENGER_TRANSPORT_DSN` is no longer supported

coldic3: I don't remove it intentionally as in lower versions of Sylius it's still needed.

Rafikooo (approved):

coldic3:
Already merged in https://github.com/Sylius/Sylius-Standard/pull/899.

PR #901, opened 2023-02-25
[Maintenance] Use PHP 8 attributes instead of annotations.

daifma:
Doctrine annotations are deprecated in favor of native PHP 8 attributes.

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -20,7 +20,7 @@ doctrine:
                 mappings:
                     App:
                         is_bundle: false
-                        type: annotation
+                        type: attribute
                         dir: '%kernel.project_dir%/src/Entity'
                         prefix: 'App\Entity'
                         alias: App
--- a/src/Entity/Addressing/Address.php
+++ b/src/Entity/Addressing/Address.php
@@ -7,10 +7,6 @@ namespace App\Entity\Addressing;
 use Doctrine\ORM\Mapping as ORM;
 use Sylius\Component\Core\Model\Address as BaseAddress;
 
-/**
- * @ORM\Entity
- * @ORM\Table(name="sylius_address")
- */
 #[ORM\Entity]
 #[ORM\Table(name: 'sylius_address')]
 class Address extends BaseAddress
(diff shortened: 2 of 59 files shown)


vvasiloi (approved):

Zales0123:
Hello @daifma and thank you for your contribution! However, we need to be careful with such changes, as they can result in potential problems for e.g. plugins creators. We already had to [revert a similar change in the past](https://github.com/Sylius/Sylius-Standard/pull/881). I believe we can move forward with such a change since 1.13 (so 1.12 would be a transitional version), but we need some other voices from the community 🚀 cc @lchrusciel and @Prometee (as you were the one that raised a concern 🖖)

Prometee:
I agree with you @Zales0123 since a transition was made in 1.12 we can go for PHP attributes with 1.13.

vvasiloi:
v1.13 branch when?

Prometee:
😄 good question, any roadmap available ?

daifma:
@Zales0123.
Attributes are only enable in src/Entity folder, 
```
 mappings:
    App:
        is_bundle: false
        type: attribute
        dir: '%kernel.project_dir%/src/Entity'
        prefix: 'App\Entity'
        alias: App
```
For the support of attributes, we already require `"php": "^8.0",` in `composer.json`.

You can close this PR if this changes had already caused problems.

Prometee:
@daifma the issue is not with Sylius-Standard, it's with all Sylius plugins which add properties or enhance Sylius entities. @Zales0123 is just warning you about the fact that this change is also affecting Plugins. Your PR is needed but can't be merged in Sylius 1.12. Your PR will be merged when Sylius 1.13 will be available as a new branch.

In your own projects, nothing stop you to use PHP attributes only, you will just have to rewrite entities plugins annotations, here are few examples: 

https://github.com/Setono/SyliusGiftCardPlugin/blob/0.11.x/src/Model/OrderItemUnitTrait.php#L11

https://github.com/Setono/SyliusGiftCardPlugin/blob/0.11.x/src/Model/OrderTrait.php#L18-L26

https://github.com/odiseoteam/RbacPlugin/blob/master/src/Entity/AdministrationRoleAwareTrait.php#L12-L17

https://github.com/loevgaard/SyliusBrandPlugin/blob/master/src/Model/ProductTrait.php#L11-L16

https://github.com/FLUX-SE/SyliusEUVatPlugin/blob/master/src/Entity/EuropeanChannelAwareTrait.php#L13-L19

PR #931, opened 2023-09-15
SyliusCon advert

Rafikooo:
We'd like to inform as many developers as possible about our second conference fully dedicated to Sylius

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -114,7 +114,8 @@
         "post-create-project-cmd": [
             "@php -r \"file_put_contents('.gitignore', str_replace('/composer.lock' . PHP_EOL, '', file_get_contents('.gitignore')));\"",
             "@php bin/console sylius:inform-about-gus --ansi",
-            "@php bin/console sylius:show-available-plugins --ansi"
+            "@php bin/console sylius:show-available-plugins --ansi",
+            "echo 'Join us at this year’s \uD83E\uDDA2 SyliusCon \uD83E\uDDA2: https://sylius.com/conference'"
         ],
         "auto-scripts": {
             "cache:clear": "symfony-cmd",


jakubtobiasz (approved):

composer.json
```diff
@@ -24,7 +24,7 @@
         }
     ],
     "require": {
-        "php": "^8.0",
+        "php": "^8.1",
```
loic425: that's strange to upgrade php version on 1.12 branch?

jakubtobiasz: https://github.com/Sylius/Sylius-Standard/pull/932#issuecomment-1724887847

Sylius-Standard doesn't follow SemVer (or at least it doesn't have to).

TheMilek (approved):

composer.json
jakubtobiasz: Anyway, our new teammate @Wojdylak will check what's wrong with 8.0 :p. So we'll wait with merging for results from his investigation.

loic425: > [#932 (comment)](https://github.com/Sylius/Sylius-Standard/pull/932#issuecomment-1724887847)
> 
> Sylius-Standard doesn't follow SemVer (or at least it doesn't have to).

Yeah indeed, I thought it was on sylius/sylius, misread

Wojdylak: In php 8.0, the sylius/sylius package installs in version 1.12.9, and there is an issue with doctrine/orm
image

Rafikooo: Could you add a conflict / write a version of the doctrine package?

jakubtobiasz: #933 I'm on it

jakubtobiasz:
Thank you, @Rafikooo!

PR #932, opened 2023-09-18
[CI][Maintenance] Remove PHP 8.0 and add PHP 8.2 to workflow, Bump minimum PHP version to 8.1

Rafikooo:
PHP 8.0 is not actively supported and its security support will end in two months (25 Nov 2023).
It's time to move forward

Ref:
https://www.php.net/supported-versions.php

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: ["8.0", "8.1"]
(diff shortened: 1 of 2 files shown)


jakubtobiasz:
As it's just a project template and the PHP version can be easily changed – I'm 👌🏻 with this change.

jakubtobiasz (approved):

diimpp (approved):

jakubtobiasz:
I guess we can close it @Rafikooo

PR #935, opened 2023-09-30
Add new Docker concept

piotrkardasz:
The new concept allows to setup environment with just one command 🚀 

```bash
make init # initialize Sylius installation with required dependencies

ENV=test make run # to change the environment to test

make debug # to run php container with xDebug + mailhog
```

### Requirements
- Docker
- Makefile

### used dependencies 
- https://github.com/sylius/docker-images - ready to use Sylius images
- https://github.com/boxboat/fixuid - fixes user ID between local machine and docker container (required for Linux)


.docker/ports.yaml
jakubtobiasz, line 1: As you introduced `.override.` file, I believe we can put these ports there and let the developers configure them on their own. As they would be changed in the `.override.` file, they won't be pushed to the repo.

compose.override.dist.yml
```diff
@@ -0,0 +1,49 @@
+version: "3.8"
+services:
+  php:
+    image: piotrkardasz/sylius-php:8.2-fixuid-alpine
```
jakubtobiasz: 1. I recommend considering whether we put Dockerfiles directly in the repo, so the developer can override it **or** create a `Dockerfile` inside the repository and extend your Dockerfile in it.
2. Did you consider using FrankenPHP? It's pretty cool (I base my images on it) and might bring us a marketing value. Kevin is working on it hard, and they have pretty good docs, also describing how to use FrankenPHP in production.

.docker/ports.yaml
piotrkardasz: Sure, it might be 👍

compose.override.dist.yml
diimpp: `Dockerfile` is definitely a must for me for `Sylius-Standard`, as end-user should be able to adjust his/her dev environment.

.docker/ports.yaml
piotrkardasz: Yes, but the default ports are already pre-configured so if there is a need you can comment or remove imports. Of course, I can remove it and squash it to `compose.override.dist.yml`

compose.override.dist.yml
piotrkardasz: 1. I would disagree, because you don't need to be an export to hijack the service configuration. If you are advanced in this topic you can always create your own dockerfile and tune the base configuration or even create a dedicated one. (The trick with an image is that you can quickly inject another image in the fly e.g. with xdebug to avoid wasting time on rebuilding the layer of the image and run the project in minutes). 
3. Yes, I have tested the pre-stable version and now I know there is a stable release. We can create an alternative configuration if needed, but in my opinion, I would go with the services configuration that everyone knows. I want to keep the configuration simple to not create a Swiss army knife.

jakubtobiasz: I'd be in favor of going full into FrankenPHP.
- there's one container (frankenphp) instead of two (nginx + php-fpm)
- it's simpler to use it
- it's production ready with pretty good documentation and examples
- it seems the community is going to adapt it, and it'd be cool to help FrankenPHP grow

piotrkardasz: FrankenPHP is nice to have 🚀 but IMHO not needed for typical Sylius development. I would stay with the current level of complexity to not add another dependency that is still not well known and requires additional understanding of how HTTP requests are handled + additional Symfony Runtime setup is needed.
If we really want to have FrankenPHP onboard Sylius, I would prefer to have an optional configuration or a guide on how to set it up with Sylius.

jakubtobiasz: > If we really want to have FrankenPHP onboard Sylius, I would prefer to have an optional configuration or a guide on how to set it up with Sylius.

Np, we can add it in our new docs we're working on.

compose.yml
```diff
@@ -0,0 +1,17 @@
+services:
+  php:
+    image: ghcr.io/sylius/sylius-php:8.2-alpine
+  mysql:
+    image: mysql:5.7 # Sylius is fully working on mysql 8.0 version
```
jakubtobiasz:
```suggestion
    image: mysql:8.0 # Sylius is fully working on mysql 8.0 version
```

AFAIR we're bumping the minimal mysql version in 1.13, previous versions are working with `8.0` so we can safely use it.

compose.override.dist.yml
```diff
@@ -0,0 +1,56 @@
...
+      APP_ENV: ${ENV:-prod}
+      APP_SECRET: EDITME
+      DATABASE_URL: "mysql://root@mysql/sylius_%kernel.environment%"
+      MAILER_URL: null://null
+      MESSENGER_TRANSPORT_DSN: doctrine://default
+      SYLIUS_MESSENGER_TRANSPORT_MAIN_DSN: doctrine://default
+      SYLIUS_MESSENGER_TRANSPORT_MAIN_FAILED_DSN: doctrine://default?queue_name=main_failed
+      SYLIUS_MESSENGER_TRANSPORT_CATALOG_PROMOTION_REMOVAL_DSN: doctrine://default?queue_name=catalog_promotion_removal
+      SYLIUS_MESSENGER_TRANSPORT_CATALOG_PROMOTION_REMOVAL_FAILED_DSN: doctrine://default?queue_name=catalog_promotion_removal_failed
+      PHP_DATE_TIMEZONE: ${PHP_DATE_TIMEZONE:-UTC}
```
jakubtobiasz:
```
        env_file:
            - .env.local
```

We can create an example `.env.local` and add it to the gitignore, as we de with `.override.dist.yml`

compose.override.dist.yml
```diff
@@ -0,0 +1,56 @@
...
+  nginx:
+    volumes:
+      - ./public:/srv/sylius/public:ro
+      # if you develop on Linux, you may use a bind-mounted host directory instead
+#      - ./public/media:/srv/sylius/public/media:ro
+      - public-media:/srv/sylius/public/media:ro,nocopy
+    ports:
+      - "80:80"
+  nodejs:
+    image: node:${NODE_VERSION:-16}-alpine
```
jakubtobiasz: It should be `20` by default.

compose.debug.yml
jakubtobiasz, line 1: XDebug has been improved, and it doesn't affect the performance if not triggered. I'd be in favor of not adding another compose file just for debugging.

Makefile
```diff
@@ -0,0 +1,40 @@
...
+	@ENV=$(ENV) DOCKER_USER=$(DOCKER_USER) docker-compose down -v
+
+php-shell:
+	@ENV=$(ENV) DOCKER_USER=$(DOCKER_USER) docker-compose exec php sh
+
+node-shell:
+	@ENV=$(ENV) DOCKER_USER=$(DOCKER_USER) docker-compose run --rm -i nodejs sh
+
+yarn-watch:
+	@ENV=$(ENV) DOCKER_USER=$(DOCKER_USER) docker-compose run --rm -i nodejs "yarn watch"
```
jakubtobiasz: We shouldn't use `yarn`, as I'm planning to push `npm` along with Sylius 2. We can start using the target stack here.

jakubtobiasz (commented):
IMO there's a lot of things happening in the compose files. Maybe we could introduce some "abstraction" by using custom Dockerfiles. For example, my compose files I'm using are +/- like this:

```
services:
    php:
        extends:
            file: /Users/jacob/Docker/compose/php.yaml
            service: frankenphp
        env_file:
            - .env.local
        environment:
            APP_URL: sylius.orb.local
        labels:
            - dev.orbstack.domains=sylius.orb.local
        volumes:
            - app:/app:delegated
        depends_on:
            - db
    db:
        extends:
            file: /Users/jacob/Docker/compose/database.yaml
            service: mariadb
        volumes:
            - database:/var/lib/mysql
        labels:
            - dev.orbstack.domains=db-sylius.orb.local
    frontend:
        extends:
            file: /Users/jacob/Docker/compose/common.yaml
            service: node
        volumes:
            - app:/app
    mailhog:
        image: mailhog/mailhog
        labels:
            - dev.orbstack.domains=mailhog-sylius.orb.local
```

Some things are due to the fact I'm using the Orbstack, but we should definitely consider making the Docker experience as simple as we can. 

From our experience, not every person touching Sylius with Docker is familiar with it.

compose.override.dist.yml
Wojdylak: I believe it's better to keep all environment variables in one place, without separating them into another file or using an example file. I have added a comment suggesting that if somebody wishes, they can move these environment variables to `.env.local`.

compose.override.dist.yml
```diff
@@ -0,0 +1,71 @@
+version: "3.8"
+services:
+    php:
+        image: ghcr.io/sylius/sylius-php:8.2-fixuid-xdebug-alpine
```
vasilvestre, line 4: Using an exterior image requires some customisation documentation

vasilvestre:
It's my own opinion but I'm not fan of the external tool dependency (fixuid) which isn't even 1.0 released.

piotrkardasz:
> It's my own opinion but I'm not fan of the external tool dependency (fixuid) which isn't even 1.0 released.

fixuid is reasonable way for DX development to handle correct UIDs and GIDs for Linux/Windows(WSL); for Mac it is redundant. 
⚠️ [Fixuid](https://github.com/boxboat/fixuid) should be used only for development purposes

jakubtobiasz (approved):

PR #942, opened 2024-01-25
Add platform.sh support

senghe:
Hi. I've already prepared a platform.sh setup configuration for Sylius. In a nearest future I'll push also a documentation for this feature into Sylius repository.

I hope you like it :vulcan_salute:


.env.platform.dist
```diff
@@ -0,0 +1 @@
+DATABASE_URL=mysql://root@127.0.0.1/sylius_%kernel.environment%?charset=utf8mb4&serverVersion=mariadb-11.0.0
```
diimpp: Is it necessary to have independent .env template for psh?

diimpp: `serverVersion` doesn't match selected PSH database.

.node-version
```diff
@@ -0,0 +1 @@
+16.20.2
```
diimpp: Unless there is reason to have minimal minor version, better to select major version only to automatically get latest version.

```suggestion
16
```

diimpp: I think 1.12 still on 16, but 1.13 on 18/20? :thinking:

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
+name: sylius
+
+type: php:8.1
```
diimpp: I think it should be either minimal (8.0 for 1.12) version or latest version.

```suggestion
type: php:8.3
```

diimpp: Build flavor should be specified at the top.

```suggestion
type: php:8.1
build:
    flavor: none
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+relationships:
+    database: "db:mysql"
+
+disk: 2048
+
+mounts:
+    "/var/cache": "shared:files/cache"
```
diimpp: Missing `config/jwt`, otherwise it will fail to generate it on readonly FS.

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+    "/var/log": "shared:files/log"
```
diimpp: To my knowledge, `shared:files/log` is outdated syntax, that can be seen in years old examples, but it's not a default syntax in docs anymore. https://docs.platform.sh/add-services/network-storage.html#2-add-the-relationship

```suggestion
    "/var/log": 
       source: local
       source_path: log
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+    "/var/sessions": "shared:files/sessions"
+    "/public/uploads": "shared:files/uploads"
```
diimpp: Missing `'/public/bundles'` unless it somehow transferred at build step.

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+    
+    deploy: |
+        set -e
+        symfony-deploy
+        rm -rf var/cache/*
+        mkdir -p public/media/image
+        bin/console sylius:install -n
+        bin/console sylius:fixtures:load -n
+        bin/console assets:install --symlink --relative public
+        bin/console cache:clear
```
diimpp, line 112: Missing 
> bin/console lexik:jwt:generate-keypair --skip-if-exists

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+hooks:
+    build: |
+        set -e 
+        cp .env.platform.dist .env.local
+        composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader
+        curl -fs https://get.symfony.com/cloud/configurator | bash
+        n auto
+        hash -r
+        yarn install
+        GULP_ENV=prod yarn build
```
diimpp: I think 1.12 is using webpack by default?
```suggestion
        yarn install --frozen-lockfile
        yarn build:prod
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+        cmd: "php bin/console sylius:promotion:generate-coupons"
+    php bin/console sylius:remove-expired-carts:
+        spec: "0 2 * * *"
+        cmd: "php bin/console sylius:cancel-unpaid-orders"
+        
+hooks:
+    build: |
+        set -e 
+        cp .env.platform.dist .env.local
+        composer --no-ansi --no-interaction install --no-progress --prefer-dist --optimize-autoloader
```
diimpp: Why not use composer build flavor?

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+        yarn install
+        GULP_ENV=prod yarn build
+    
+    deploy: |
+        set -e
+        symfony-deploy
+        rm -rf var/cache/*
+        mkdir -p public/media/image
+        bin/console sylius:install -n
+        bin/console sylius:fixtures:load -n
```
diimpp:
```suggestion
        bin/console sylius:install -n
        bin/console sylius:fixtures:load -n
        # {Smart comment placeholder :') }
        # doctrine:migration:migrate --no-interaction
```

.platform.app.yaml
diimpp: Overall placement feels weird, I personally would go for

- type
- build
- runtime
- variables
- dependencies
- hooks
- relationships
- disk
- mounts
- web
- workers
- cron

or last three as
- workers
- cron
- web

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+web:
+    locations:
+        "/":
+            root: "public"
+            passthru: "/index.php"
+            allow: true
+            expires: -1
+            scripts: true
+        '/assets/shop':
```
diimpp, line 44: For webpack,
```suggestion
        '/build':
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+        N_PREFIX: /app/.global
+
+build:
+    flavor: none
+
+dependencies:
+    nodejs:
+        n: "*"
+        yarn: "*"
+        gulp-cli: "*"
```
diimpp:
```suggestion
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+    php:
+        composer/composer: '^2'
+
+crons:
+    cancel-unpaid-orders:
+        spec: "0 2 * * *"
+        cmd: "php bin/console sylius:cancel-unpaid-orders"
+    generate-promotion-coupons:
+        spec: "0 2 * * *"
+        cmd: "php bin/console sylius:promotion:generate-coupons"
```
diimpp: Unless I'm mistaken, this command can't be used like so and needs to be removed.

image

```suggestion
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+dependencies:
+    nodejs:
+        n: "*"
+        yarn: "*"
+        gulp-cli: "*"
+    php:
+        composer/composer: '^2'
+
+crons:
+    cancel-unpaid-orders:
```
diimpp: `:` is not a valid char at worker name (even wrapped on ' '), but can be used at cron, but it doesn't sound like a good idea take advantage of this bug.

```suggestion
    'sylius:cancel-unpaid-orders':
```
```suggestion
    sylius-cancel-unpaid-orders:
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+        spec: "0 2 * * *"
+        cmd: "php bin/console sylius:cancel-unpaid-orders"
+    generate-promotion-coupons:
+        spec: "0 2 * * *"
+        cmd: "php bin/console sylius:promotion:generate-coupons"
+    php bin/console sylius:remove-expired-carts:
```
diimpp:
```suggestion
    sylius-remove-expired-carts:
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+        spec: "0 2 * * *"
+        cmd: "php bin/console sylius:cancel-unpaid-orders"
```
diimpp: Not sure, that cron commands get `APP_ENV=prod`, better be safe, than sorry :smiley: 

```suggestion
        cmd: "php bin/console --env=prod sylius:cancel-unpaid-orders"
```

.platform.app.yaml
```diff
@@ -0,0 +1,99 @@
...
+        n: "*"
+        yarn: "*"
+        gulp-cli: "*"
+    php:
+        composer/composer: '^2'
+
+crons:
+    cancel-unpaid-orders:
+        spec: "0 2 * * *"
+        cmd: "php bin/console sylius:cancel-unpaid-orders"
```
diimpp:
```suggestion
        cmd: "php bin/console --env=prod sylius:cancel-unpaid-orders"
```

diimpp (dismissed):

.platform.app.yaml
diimpp: Personally I would remove `files` part from `source_path` as each definition is a separate mount and there is no point prefixing them.

.platform.app.yaml
TheMilek: I will leave it for now as it is similar to config provided for the template by platform.sh 🤔 https://github.com/platformsh-templates/sylius/blob/master/.platform.app.yaml

.platform.app.yaml
TheMilek: seems as it is not needed, works without it
edit: nevermind you were right with the bundles, the assets were not installing 😅

mpysiak (approved):

GSadee (approved):

GSadee:
Thank you, Marcin! :1st_place_medal:

PR #958, opened 2024-03-26
SYL-3329 Install PayPalPlugin

mpysiak:


config/packages/http_discovery.yaml
```diff
@@ -0,0 +1,10 @@
+services:
+    Psr\Http\Message\RequestFactoryInterface: '@http_discovery.psr17_factory'
```
Wojdylak: Why do we need to add this file?

mpysiak: Some file left from flex. Removed

Wojdylak (approved):

Wojdylak:
Thank you, Michał :tada:

PR #975, opened 2024-04-23
Fix error during template install

frickelbruder:
Fixes bug during installation and enables the missing "sylius_abstraction.state_machine" service.

During installation of a clean Sylius using 
`composer create-project sylius/sylius-standard MyFirstShop`
an error occurs because of the missing sylius_abstraction.state_machine service definition.

```
!!  In CheckExceptionOnInvalidReferenceBehaviorPass.php line 119:
!!                                                                                                                       
!!    The service "sylius.fixture.order" has a dependency on a non-existent service "sylius_abstraction.state_machine". 
```

This MR fixes this behaviour.


lchrusciel (approved):

GSadee:
The above problem should already be resolved in two ways: we have added a recipe to add this bundle to bundles when it is installed, additionally it has been added to the bundles in Sylius-Standard 1.13 and a new version of our tempalte has been released, which installs 1.13 by default instead of 1.12.

PR #982, opened 2024-04-29
[Maintenance] Require precise packages instead of generic sylius/sylius

lchrusciel:
I propose switching from the generic "sylius/sylius" dependency to a set of high-level, more explicit ones. Of course, I would like to require "sylius/core-bundle" - as it is the core and heart of Sylius itself. In addition, with this change, we will require:

*  "sylius/admin-bundle"
* "sylius/shop-bundle" 
* "sylius/api-bundle"

These packages should be considered as a default stack of Sylius at the moment. What is missing from this list? Deprecated admin-api, but that should be considered as a benefit of this change.

In addition what we are earning with it is the fact that if in the future we will add a new layer to Sylius it won't be automatically propagated to all existing shops. Also, we are making it easier for people to turn off one of the accessing layers to Sylius

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,9 @@
     "require": {
         "php": "^8.1",
         "sylius/paypal-plugin": "^1.6",
-        "sylius/sylius": "~1.13.0",
+        "sylius/admin-bundle": "~1.13.0",
+        "sylius/shop-bundle": "~1.13.0",
+        "sylius/api-bundle": "~1.13.0",
         "symfony/dotenv": "^5.4 || ^6.4",
         "symfony/flex": "^2.4",
         "symfony/runtime": "^5.4 || ^6.4"


jakubtobiasz (approved):

diimpp (approved):
Isn't core automatically required by admin, shop, api? Is it necessary to have explicit requirement?

composer.json
```diff
@@ -26,7 +26,10 @@
     "require": {
         "php": "^8.1",
         "sylius/paypal-plugin": "^1.6",
-        "sylius/sylius": "~1.13.0",
+        "sylius/core-bundle": "~1.13.0",
```
Prometee:
```suggestion
```

Like @diimpp mentioned it, core is already required by shop, admin or api bundle. Then it could be omitted.

Prometee (changes requested):

composer.json
lchrusciel: To be honest, yes it is not required. However, I think it is good practice to have explicit composer requirements for the things that you use. Now, we can argue that the same is true for all the other sub-dependencies of sylius/core-bundle. And here I would go pragmatic and just assume that core-bundle is an aggregator for all of them. 

To sum up, I guess we would be fine with leaving core-bundle as it is. But removing it will also not change too much

Prometee (approved):

Rafikooo:
We’re going to process this in the near future, but for now I’ll close the ticket.

PR #983, opened 2024-04-29
[Maintenance] Bump PHP and Symfony dependencies

lchrusciel:
Hello there!

My next proposal concerns bumping requirements in Sylius-Standard. I propose requiring only the latest version of PHP and Symfony for the current version. There is no need to require a lower version of these packages. No one would like to use older Symfony LTS just for the sake of it. And with Flex as a hard dependency, we are installing Symfony 6.4 anyway. 

With PHP it is almost the same. Plus, a composer will sort out which version of Sylius-Standard it should use on its own

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,12 @@
         }
     ],
     "require": {
-        "php": "^8.1",
+        "php": "^8.3",
         "sylius/paypal-plugin": "^1.6",
         "sylius/sylius": "~1.13.0",
-        "symfony/dotenv": "^5.4 || ^6.4",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -83,7 +83,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-main": "1.13-dev"
+            "dev-main": "1.14-dev"
```
GSadee, line 86: It should still be 1.13-dev
```suggestion
            "dev-main": "1.13-dev"
```

PR #994, opened 2024-05-22
[Maintenance] Add Sylius Plus Info command

mpysiak:
console view after creating new project with composer:

Screenshot 2024-05-22 at 10 00 59

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "require": {
         "php": "^8.0",
         "sylius/paypal-plugin": "^1.5.2",
-        "sylius/sylius": "~1.12.11",
+        "sylius/sylius": "~1.12.17",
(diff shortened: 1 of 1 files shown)


GSadee (commented):
The command and its output looks ok 💪🏻, but after thinking about it and discussing it internally, I would be in favour of moving it to `Sylius/Sylius` after all. Mainly so that when users set up a new project, they do not have an unnecessary command at the start. In addition, we may yet use it in Sylius itself 🤔.

GSadee (approved):

GSadee:
However, this PR should wait for the next Sylius releases and we should bump up Sylius version in composer

Wojdylak (approved):

GSadee:
Thanks, Michał! :tada:

PR #999, opened 2024-06-07
Cart quantity field

azy64:
# In that pull request I have:

- Fork and Clone the Sylius Standard Repository

- Modify Cart Quantity Field:
 - Adjust the "quantity" field in the cart forms on the product pages.
- Change the step increment from 1 to 10, starting at 10.

This means the quantities available should be 10, 20, 30, and so on, instead of the default 1, 2, 3, etc.

- Additional Task (Optional):
  1. If the user selects a quantity of 70, display a plain JavaScript alert with the message "Great Choice!".

<b>NB</b>: There are many solutions for this problem, but I've chosen to override a specific template and put a javascript code, thus I'll not modify many templates either.

### DIFF
--- a/.env
+++ b/.env
@@ -10,7 +10,8 @@ APP_SECRET=EDITME
 
 ###> doctrine/doctrine-bundle ###
 # Choose one of the following DBMS, adjust the server version and charset if needed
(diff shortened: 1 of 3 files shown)


GSadee:
Hello @azy64!

I believe that the solution you wanted to achieve should only be implemented in your end application and not in the Sylius project template, as this is a very specific case 🖖🏻

PR #1002, opened 2024-06-17
[Composer] Bump up PayPalPlugin dependency

GSadee:
As it is a template for creating a Sylius project, I think it would be good to encourage the use of the latest versions of our dependencies. So I would like to propose bumping up the PayPalPlugin to its latest minor version which also clarifies some other dependencies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,14 +27,10 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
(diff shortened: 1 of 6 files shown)


NoResponseMate (approved):

Wojdylak (approved):

Wojdylak:
Thank you, @GSadee :tada:

PR #1028, opened 2024-09-25
[Maintenance] Upgrade Recipes & Refine Project Structure

Rafikooo:
This PR refreshes the project structure by upgrading Symfony recipes, committing default bundle configurations, and removing obsolete code.

Key changes:
	1.	Dropped PHPSpec
	2.	Adjusted PHPStan and PHPUnit configurations
	3.	Updated third-party recipe configuration files to align with Sylius
	4.	Removed annotations
	5.	Moved the migrations directory

An example migration:
image

### DIFF
--- a/config/packages/nyholm_psr7.yaml
+++ b/config/packages/nyholm_psr7.yaml
@@ -1,0 +1,11 @@
+services:
+    # Register nyholm/psr7 services for autowiring with PSR-17 (HTTP factories)
+    Psr\Http\Message\RequestFactoryInterface: '@nyholm.psr7.psr17_factory'
+    Psr\Http\Message\ResponseFactoryInterface: '@nyholm.psr7.psr17_factory'
+    Psr\Http\Message\ServerRequestFactoryInterface: '@nyholm.psr7.psr17_factory'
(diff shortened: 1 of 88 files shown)


mpysiak (approved):

symfony.lock
```diff
@@ -525,274 +271,179 @@
...
-    },
-    "symfony/thanks": {
-        "version": "v1.0.8"
+            "version": "2.13",
+            "ref": "6acd9ff4f7fd5626d2962109bd4ebab351d43c43"
+        },
+        "files": [
+            "assets/bootstrap.js",
+            "assets/controllers.json",
+            "assets/controllers/hello_controller.js"
```
Rafikooo, line 416: I was unsure how to handle these files and decided to remove them, as the way we have configured our UI seems to follow a different architecture

Wojdylak (approved):

config/packages/nyholm_psr7.yaml
GSadee, line 1: I would be for removing this config as nyholm is a dev dependency and probably we should not force to have it in the config in the end application

GSadee (approved):

PR #1060, opened 2024-10-31
Improve configurations and routes

GSadee:


Rafikooo (approved):

quentint:
In `config/packages/sylius_api.yaml` I see this:

```yaml
when@dev:
    sylius_api:
        enabled: true
```

This makes the API available in `dev` mode, but not in `prod` env. I'm not sure I understand this choice, can you tell me more?

I'm working on an app, relying on the API (with custom read-only endpoints). When switching to the `prod` env, my app broke, and I couldn't figure out why. I had to dig a bit, and found this 😉

Really interested in hearing your motivations!
Thanks!

PR #1063, opened 2024-11-04
Add missing entities to 2.0

GSadee:
This PR is waiting for the next version of Sylius (v2.0.0-RC.1)


mpysiak (approved):

src/Entity/User/AvatarImage.php
diimpp, line 1: Personally I would have put it at `Entity\AdminUser\AvatarImage`, but such separation doesn't exist in the main package.

diimpp (approved):

PR #1072, opened 2024-11-14
fix: make init error with docker check

### LINKED ISSUE #1018: Incorrect check in Makefile
In the Makefile file, there's a check that checks if docker exists, but the issue is that you used `DOCKER_COMPOSE` variable 
```
@which $(DOCKER_COMPOSE) > /dev/null || (echo "Please install docker compose binary" && exit 1)
```
and `DOCKER_COMPOSE` variable consists of `docker compose` which in turn causes `which docker compose` command, and while docker command exists, compose command does not exist because its parameter to the docker command, so the check fails and currently `make init` will always fail on every system unless they somehow got command named `compose`

sonbui00:
FIx issue when run `make init`

`make[1]: *** [docker-compose-check] Error 1`

Fix #1018

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,6 @@ node-watch:
 	@ENV=$(ENV) DOCKER_USER=$(DOCKER_USER) $(DOCKER_COMPOSE) run --rm -i nodejs "npm run watch"
 
 docker-compose-check:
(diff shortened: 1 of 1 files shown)


Wojdylak (approved):

GSadee:
@sonbui00, thank you for the contribution! Could you maybe open the same fix for the 1.14 branch?

PR #1075, opened 2024-11-20
[Maintenance] Add app webpack endpoint to twig hooks

Wojdylak:


assets/admin/entry.js
```diff
@@ -1 +1,2 @@
 // In this file you can import assets like images or stylesheets
+console.log('Hello Webpack Encore! Edit me in assets/shop/entry.js');
```
loic425: ??
```suggestion
console.log('Hello Webpack Encore! Edit me in assets/admin/entry.js');
```

mpysiak (approved):

templates/admin/javascripts.html.twig
```diff
@@ -0,0 +1,2 @@
+
```
GSadee, line 1:
```suggestion
```

GSadee (approved):

PR #1077, opened 2024-11-24
Minimum quantity 10 of the add to cart page and steps change to 10

priyank780:


diimpp:
Hi @priyank780 , I assume you have opened this PR by error from your project.

But just on case, such change cannot be accepted into master as it's not usable by anyone outside of your project

PR #1079, opened 2024-11-25
Fix docker nodejs permission in MacOS

sonbui00:
There is an error when run `make init` (v2.0) in MacOS with nodejs container


rust-le:

PR #1084, opened 2024-12-03
change quantity step to 10

galion96:


GSadee:
I believe that this is not the right repository to which you have opened this PR 🖖🏻

PR #1091, opened 2024-12-11
[Maintenance] New `symfony/form` Recipe

Rafikooo:


GSadee:
This is the recipe and config for Symfony 7.2, I'm not sure if we should merge it to the 2.0 branch 🤔

PR #1117, opened 2025-04-10
Add Sylius/MolliePlugin as a required dependency

TheMilek:


config/routes/sylius_mollie.yaml
```diff
@@ -0,0 +1,2 @@
+sylius_mollie_plugin:
+    resource: "@SyliusMolliePlugin/config/routing.yaml"
```
Rafikooo:
```suggestion
    resource: "@SyliusMolliePlugin/config/routes.yaml"
```

behat.yml.dist
```diff
@@ -61,3 +61,69 @@ default:
     gherkin:
         filters:
             tags: "~@todo&&~@cli" # CLI is excluded as it registers an error handler that mutes fatal errors
+
+
```
GSadee, line 65:
```suggestion

```

Wojdylak (approved):

GSadee (approved):

Rafikooo (approved):

oallain:
Hello @TheMilek ,

Can you add a setting to disable it easily ?
Because a lot of people are going to have to make the exact opposite changes to yours on their projects if they don't want to use Mollie.

Thanks :)

jacquesbh:
Hi!

I agree with @oallain! Adding Mollie by default on Sylius is a bad move IMO.

We will do exactly as Olivier said, doing a revert of those changes, making it more complicated to install Sylius with the Sylius-Standard.

Why adding such plugins by default?

maximehuran:
We used to remove Paypal because it's not necessary on a Sylius Standard project.
Now we will have to remove Mollie too.

It's the opposite of the Sylius philosophy to force people to have these plugins.
Mollie force users to have Invoicing and Refund plugin and a lot of things which is not needed by a lot of projects.

Prometee:
@GSadee I agree too, why this is not part of the Sylius demo instead ?
This plugins is not qualify to be shipped inside this project like it is developed today.

vasilvestre:
Such changes **should** have an issue linked, maybe a decision record and at least a description.

CoderMaggie:
Hi everyone,
Thank you for your valuable feedback and for engaging in this discussion.

Adding the Mollie plugin to Sylius-Standard is a strategic decision that is aligned with our official partnership, just like PayPal, which has been part of the Standard for some time now.

What’s crucial here is that the **Mollie plugin is now fully maintained by the Sylius team**. Thanks to the dedicated resources from this collaboration, we’ve already delivered a major refactor, continue to improve the plugin actively, and have more enhancements on the roadmap, including full Sylius 2.0 support. This ensures long-term stability and seamless integration, offering the community a reliable, first-class payment solution.

Including key payment plugins by default serves two main goals:
- Boosting adoption of well-maintained, trusted solutions.
- Providing a fully functional starting point in Sylius-Standard, especially for agencies and merchants who expect essential features out of the box.

We’re aware that some projects prefer a minimal setup, as flexibility has always been part of Sylius' DNA. That’s why we’re reviewing this PR again to make sure disabling Mollie remains simple for those who don’t need it. 🕵🏻‍♀️

👀 At the same time, I’d like to hint that we’re already working towards a future where Sylius will offer even more modularity, similar to how Symfony allows you to pick exactly what you need. This will enable predefined setups for faster project kick-offs (e.g., payments, CMS, etc.), which is especially useful for agencies. Until then, Sylius-Standard will continue to include essential extensions like payment gateways to ensure a smooth experience for most use cases.
Both Mollie and PayPal will also be showcased in the Sylius Demo to present their capabilities right away.

We believe this approach strengthens the ecosystem by combining quality, stability, and developer choice, today and even more so in the near future.

Thanks again for contributing to this conversation!

jacquesbh:
Hi @CoderMaggie !

This PR should be reverted in the meanwhile. We need to be able to disable this plugin and all the requirements it adds (invoicing, refund…) before it is merged to Sylius-Standard.

This kind of move kills all CIs of all plugins out there.

**It is a bad move.**

I understands that Strategy has to move on, but if it's on despite of everything else it seems to me that this should be reconsidered.

All the best, as always.

oallain:
Thank you @CoderMaggie for your message.

I understand your desire to improve and I have nothing against Mollie.
My problem is that in addition to everything that has already been mentioned in previous messages, we are forcing a technical debt into a Sylius project.
Let me explain:
`sylius/sylius-standard` require `sylius/mollie-plugin`
`sylius/mollie-plugin` require `sylius/refund-plugin`
`sylius/refund-plugin` require wkhtmltopdf

And wkhtmltopdf has been abandoned for years ... and hosting companies no longer want to install it. 
You're going to tell me that PDF generation can be disabled, but it's enabled by default.

NB: This problem could be solved by using sensiolabs/gotenberg-bundle

Have a nice day !

mpysiak:
Hi @jacquesbh,

Out of curiosity—could you explain how you’re using Sylius Standard in your CI setup? 🖖

maximehuran:
Hi @mpysiak,

Our CI run `make install` which run a lot of things in our Makefile like the `composer create-porject` of `sylius-standard`

https://github.com/monsieurbiz/SyliusCmsPagePlugin/blob/master/Makefile#L67

jacquesbh:
> Hi @jacquesbh,
> 
> Out of curiosity—could you explain how you’re using Sylius Standard in your CI setup? 🖖

Hi @mpysiak!

The Mollie fixtures add a new Channel. Which means that Sylius Standard comes with 2 channels, meaning we can't access the First Channel just by going to the URL serving the project. We need to make an update in our CIs or installation processes to update the URL of those channels to avoid a 500 by default.

jacquesbh:
> Thank you @CoderMaggie for your message.
> 
> I understand your desire to improve and I have nothing against Mollie. My problem is that in addition to everything that has already been mentioned in previous messages, we are forcing a technical debt into a Sylius project. Let me explain: `sylius/sylius-standard` require `sylius/mollie-plugin` `sylius/mollie-plugin` require `sylius/refund-plugin` `sylius/refund-plugin` require wkhtmltopdf
> 
> And wkhtmltopdf has been abandoned for years ... and hosting companies no longer want to install it. You're going to tell me that PDF generation can be disabled, but it's enabled by default.
> 
> NB: This problem could be solved by using sensiolabs/gotenberg-bundle
> 
> Have a nice day !

And even… We don't always want the refund and invoicing plugins in our projects!
Actually that is a key point of Sylius! Its core is lightweight, it doesn't include potential useless features.
The Mollie plugin, even maintained by Sylius itself, is going in the opposite direction by adding all those requirements.
And all the changes to the entities, to the frontend, to the admin panel (by adding some special tags in all admin pages just for one plugin in some specific pages)…

Removing it will lead to forking the Sylius-Standard, which is a bad alternative, IMO.

bitbager:
Hi everyone!

Mollie is one of the most involved PSPs we have worked with for many years now. You all know they're super close to our dev communities. But I understand this might be worrying, hence here's my input:

1. +1 for the wkhtmltopdf dependency, we should have that removed/replaced with something more up to date.
2. +1 for making it easy to disable
3. +1 for having the dependency softer in Sylius-Standard, and making Sylius-Standard lightweight, like Silex was to Symfony one day. In fact, we should do that to more packages.
4. +1 for the ADR of that decision, where we should probably describe how Mollie has been involved in the process, how the DX of the plugin has/will change. Quick spoiler alert - Mollie is willing to be an early adopter of Payment Request, and we really need such collaboration to make sure it works smoothly in the real world.

We will soon put all these things under serious consideration, thanks to your feedback. We definitely missed a few key points here, but don't worry. We will make up for it. Sylius won't change its DNA. We will work on a patch release to adapt to this feedback shortly :)

maximehuran:
Fresh installation of Sylius with Sylius standard is a mess since this change

## Front

image

## Admin

image

PR #1121, opened 2025-04-16
Fix twig mapping

Rafikooo:
```
sylius_twig_hooks:
    hooks:
        'sylius_shop.account.profile_update.update.content.main.form':
            phone_number:
                enabled: false
            secondary_phone_number:
                template: '@App/account/profile_update/update/content/main/form/secondary_phone_number.html.twig'
                priority: 600

```

To support binding `@App` to the `templates` directory,
this change also resolves the corresponding error:
image

### DIFF
--- a/config/packages/twig.yaml
+++ b/config/packages/twig.yaml
@@ -1,5 +1,7 @@
 twig:
-    paths: ['%kernel.project_dir%/templates']
+    default_path: '%kernel.project_dir%/templates'
+    paths:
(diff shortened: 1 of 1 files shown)


Prometee:
@Rafikooo using `@App` is not common instead you generally need to simply set the template without any namespace:

```yaml
template: 'account/profile_update/update/content/main/form/secondary_phone_number.html.twig'
```

Rafikooo:
Thanks @Prometee for letting me know :)

PR #797, opened 2022-07-15
Avoid migrations to be autoloaded

vasilvestre:
Closes #654

PR #1122, opened 2025-04-17
Avoid migrations to be autoloaded

### LINKED ISSUE #654: Avoid migrations to be autoloaded
Doctrine migration package specifies that migrations should NOT be autoloaded, thus the namespace is arbitrary and should be different from App\Migrations, which is the case in this distribution.

https://github.com/Sylius/Sylius/blob/master/UPGRADE-1.8.md#migrations

vasilvestre:
Superseed https://github.com/Sylius/Sylius-Standard/pull/797
Closes https://github.com/Sylius/Sylius-Standard/issues/654


GSadee (approved):

GSadee:
Thank you @vasilvestre! 🎉

PR #1142, opened 2025-06-04
Enable Sylius-Standard for Sylius 2.1

mpysiak:


.github/workflows/ci_static-checks.yaml
```diff
@@ -96,8 +96,8 @@ jobs:
             -   name: "Install dependencies"
                 run: composer update --no-interaction --no-scripts
 
-            -   name: "Validate composer.json"
-                run: composer validate --strict --no-check-version
+#            -   name: "Validate composer.json"
+#                run: composer validate --strict --no-check-version
```
GSadee, line 100: We should remember to revert this change

GSadee (approved):

PR #1153, opened 2025-07-22
Fix doctrine cache on production 2.1

### LINKED ISSUE #1151: Error with doctrine cache on prod env
On a new Sylius installation, with doctrine/dbal 3.10.0 and doctrine/orm 3.5.0, the doctrine/cache library will not be installed. This will lead to an error on prod env due to the missing Doctrine\Common\Cache\Psr6\DoctrineProvider class.
The first simple solution will be to add doctrine/cache to composer.json requirements.

Anyway, the doctrine/cache library is unmaintained and deprecated, so the best solution will be to update the packages/doctrine.yaml file with a new cache system.

mpysiak:
`Doctrine\Common\Cache\Psr6\DoctrineProvider` is not available any more in latest doctrine

Fixes: https://github.com/Sylius/Sylius-Standard/issues/1151

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -27,27 +27,14 @@ when@prod: &doctrine_prod
             entity_managers:
                 default:
                     metadata_cache_driver:
-                        type: service
-                        id: doctrine.system_cache_provider
+                        type: pool
+                        pool: doctrine.system_cache_pool
                     query_cache_driver:
-                        type: service
-                        id: doctrine.system_cache_provider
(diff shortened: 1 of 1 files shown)


vasilvestre (commented):
This could be a great opportunity to match the latest recipes
https://github.com/symfony/recipes/blob/main/doctrine/doctrine-bundle/2.13/config/packages/doctrine.yaml

delyriand:
Related issue #1153

TheMilek (approved):

PR #1156, opened 2025-08-07
Add field note to entity Order

michalkaczmarek1:


Prometee:
Hello @michalkaczmarek1 !

This field already exists on the order and it's a field the customer can fill during the last order checkout step.

Also note that this repository is not the one you will commit to if you want to contribute to Sylius, use https://github.com/Sylius/Sylius instead.

PR #1168, opened 2025-10-10
Add Adyen plugin

NoResponseMate:


Rafikooo:
In case it’s needed, we can reopen it 💃

PR #1175, opened 2025-10-14
Some mounts require to be shared between instances

vrobert78:
It's required that mounts containing data are shared across all the instances if autoscaling or rolling deployments are used.


Rafikooo (approved):
Could you please rebase the PR to remove the merge commit?

Rafikooo (approved):

PR #1303, opened 2026-06-26
Fix Stripe plugin recipe

loic425:


assets/admin/entrypoint.js
```diff
@@ -1,5 +1,5 @@
 import '@vendor/sylius/mollie-plugin/assets/admin/entrypoint';
-import '@vendor/flux-se/sylius-stripe-plugin/assets/admin/entrypoint';
+import '../../vendor/flux-se/sylius-stripe-plugin/assets/admin/entrypoint';
```
loic425: We need to keep the same line as the recipe, cause if we remove the plugin, this line will not be removed and that will cause a cache clear error.

symfony.lock
```diff
@@ -53,15 +53,16 @@
...
             "branch": "main",
-            "version": "1.0",
-            "ref": "2dd2498b504978c9ade500bc07f9603bc023af9d"
+            "version": "2.0",
+            "ref": "143bd4a3ffc977255dbdf24352526a6d3c341a76"
         },
         "files": [
-            "config/packages/sylius_stripe.yaml"
+            "config/packages/sylius_stripe.yaml",
+            "config/routes/sylius_stripe.yaml"
```
loic425, line 65: The recipe was not up to date, and the routing configuration file was not removed...

TheMilek (approved):

